PR #14, opened 2016-01-12
Add $schema to output

domoritz:


YousefED:
Great, thanks!

I think it would make more sense though if this line would be in the generateSchema function right?

domoritz:
> I think it would make more sense though if this line would be in the generateSchema function right?

Done

PR #20, opened 2016-02-17
Make more properties configurable, add some tests

### LINKED ISSUE #13: add tests to project
For example, start off with tests from typson: https://github.com/lbovet/typson/tree/master/test

domoritz:
Fixes #13


domoritz:
CC @mprudhom

domoritz:
@YousefED If you could add travis, the tests could run automatically.

YousefED:
Great stuff! Also thanks for making args a nicely typed object.

My only question before merging; any specific reason you've removed the "dist" outdir and opted for outputting in the root dir?

domoritz:
> My only question before merging; any specific reason you've removed the "dist" outdir and opted for outputting in the root dir?

It made some things much easier. For example if the tests are compiled in the same directory as the ts file, I can refer to the tjs.js file, and the test examples in ts and it compiles to the correct js. I'm not sure how I would have to rewrite the relative import paths.

YousefED:
Ok makes sense. Ideally I'd prefer a different output directory, but no priority to sort this out now I think

domoritz:
Yeah, I'd prefer it as well but the effort to make it work didn't seem worth it at this point.

PR #21, opened 2016-02-17
[WIP] Add test for string literals

domoritz:
cc @mprudhom if this worked, we could replace a few `@enum` in vega lite with string literals. Not a priority, though.


domoritz:
We have this test already.

PR #22, opened 2016-02-17
Fix typescript version to 1.7 for now

domoritz:
See https://github.com/vega/vega-lite/issues/1011#issuecomment-185403803


domoritz:
Let's not merge this now since @mprudhom may be looking into whether we can get things to work with ts 1.8.

PR #24, opened 2016-02-19
Tests for Array Types and Type Aliases

wenLiangcan:
For TypeScript compiler, type `[string, number]` accepts the following values:
```typescript
['a', 1]
['a', 1, 2]
['a', 1, 2, 'b']
```

But I wish such a type can represent a fixed size array, so I added test files under the folder `type-aliases-fixed-size-array`. What do you think about this idea?

### DIFF
--- a/test/programs/array-types/main.ts
+++ b/test/programs/array-types/main.ts
@@ -1,0 +1,3 @@
+interface MyArray {
+    [index: number]: string | number;
+}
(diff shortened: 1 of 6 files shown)


YousefED:
Note: I have not implemented this as fixed-size as this would not comply with the typescript spec. You should set minItems separately 

This can be done via comments, e.g.:

```
/**
* @TJS-minItems 2
*/
```

wenLiangcan:
Ok, I got it. And thank you for the great job!

PR #28, opened 2016-02-20
Generate required array for non-optional properties

### LINKED ISSUE #4: interface definition does not generate 'required' property in json-schema
With typson, properties of an interface are serialized as 'required' unless they are marked optional.

Given an interface like:
```typescript
interface Foo {
   bar? : number
   baz: string 
}
```
In the resulting json-schema, I'd expect **baz** to be marked _required_  and **bar** not. This is based on my previous experiments with typson. That said, toggling this behavior via a config setting or command line parameter would be nice.

It seems like this implementation is significantly different from typson and I maybe shouldn't assume it's a superset of typson?

mprudhom:
Fixes #4


YousefED:
Thanks! I did change generateRequired to false by default though. Due to the issue described in #4

PR #30, opened 2016-02-21
Fix typo in "required" property argument handing

mprudhom:


YousefED:
Thanks. Would love to have the args more statically typed (Settings class or something), but no prio

Issue #26, opened 2016-02-20
merge main logic for types to schema


YousefED:
some functions are not neatly organized. Reconsider:

getTypeDefinition
getDefinitionForType
getClassDefinition

Some logic here should be merged

YousefED:
closed by #34

PR #34, opened 2016-03-06
Refactor and recursion support

### LINKED ISSUE #33: Avoid infinite loops
With references we should be able to support interfaces that reference themselves, The easiest fix would be to only recurse if the type does not yet exist in the map of definitions.

YousefED:
closes #33 #26


YousefED:
closes #26

Issue #37, opened 2016-03-24
Top-level type annotations.


igrayson:
Given the file `./dimension.ts`
```
/**
 Some dimension 
 @additionalProperties false
*/
interface Dimension {
    /**  Width in cm */
        width: number;

    /** Height in cm */
        height: number;

    /** Length in cm */
        length:  number;
}
```
we observe this output:
```
$ typescript-json-schema dimension.ts Dimension
{
    "type": "object",
    "properties": {
        "width": {
            "type": "number",
            "description": "Width in cm"
        },
        "height": {
            "type": "number",
            "description": "Height in cm"
        },
        "length": {
            "type": "number",
            "description": "Length in cm"
        }
    },
    "$schema": "http://json-schema.org/draft-04/schema#"
}
```

Do we expect the top-level type to include a `description` (and `additionalProperties`) attribute?

`typson` output:
```
$ typson schema ./dimension.ts
{
  "Dimension": {
    "id": "Dimension",
    "type": "object",
    "description": "Some dimension",
    "additionalProperties": false,
    "properties": {
      "width": {
        "description": "Width in cm",
        "type": "number"
      },
      "height": {
        "description": "Height in cm",
        "type": "number"
      },
      "length": {
        "description": "Length in cm",
        "type": "number"
      }
    },
    "required": [
      "width",
      "height",
      "length"
    ]
  }
}
```

Issue #35, opened 2016-03-21
Error while trying to generate schema


digeff:
Any idea why I'm getting this:
λ typescript-json-schema myTSFile.ts MyMainType
npm/node_modules/typescript-json-schema/typescript-json-schema.js:120
                    else if (propertyType.getSymbol().getName() == "Array") {
                                                     ^

TypeError: Cannot read property 'getName' of undefined
    at JsonSchemaGenerator.getDefinitionForRootType (npm/node_modules/typescript-json-schema/typescript-json-schema.js:120:54)
    at JsonSchemaGenerator.getTypeDefinition (npm/node_modules/typescript-json-schema/typescript-json-schema.js:277:29)
    at npm/node_modules/typescript-json-schema/typescript-json-schema.js:90:34
    at Array.map (native)
    at JsonSchemaGenerator.getDefinitionForRootType (npm/node_modules/typescript-json-schema/typescript-json-schema.js:89:45)
    at JsonSchemaGenerator.getTypeDefinition (npm/node_modules/typescript-json-schema/typescript-json-schema.js:277:29)
    at JsonSchemaGenerator.getDefinitionForProperty (npm/node_modules/typescript-json-schema/typescript-json-schema.js:135:35)
    at npm/node_modules/typescript-json-schema/typescript-json-schema.js:237:41
    at Array.reduce (native)
    at JsonSchemaGenerator.getClassDefinition (npm/node_modules/typescript-json-schema/typescript-json-schema.js:235:49)

YousefED:
Can you provide your ts file? Thanks!

digeff:
`// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT license. See LICENSE file in the project root for details.

/* Event types used by the ProcessExecutionRecorder and ProcessExecutionSimulator */

export interface ITimedEvent {
    after: number;
}

export interface IStdOutEvent extends ITimedEvent {
    stdout: { data: string };
}

export interface IStdErrEvent extends ITimedEvent {
    stderr: { data: string };
}

export interface IErrorEvent extends ITimedEvent {
    error: { error: any };
}

export interface IExitEvent extends ITimedEvent {
    exit: { code: number };
}

export interface ICustomEvent extends ITimedEvent {
    custom: { lambda: () => Q.Promise<void> | void };
}

export type IEventArguments = IStdOutEvent | IStdErrEvent | IErrorEvent | IExitEvent | ICustomEvent;

[...]

PR #41, opened 2016-04-09
A few bug fixes, and added a pair of new arguments

otgerp:
Changes:

- Fixed the generation failing if the _tsconfig.json_ file had an "_out_" option, as then Typescript aborted with: `Option 'noEmit' cannot be specified with option 'out'.`
  
- Fixed annotations in comments only being parsed for member properties, but not for e.g. top level classes (Fixes issue #37)
  
- Fixed bug with anonymous/inline types like the following example, where it was creating `Field` as `{ "$ref": "#/definitions/{ SubfieldA: number; }" }`
```
	interface TestObject {
		Field: {
			SubfieldA: number;
		}
	}
```

- Fixed crash with inline string literals like the one below. (this should fix issue #35)
```
	type SomeEnumType = "ValueA" | "ValueB" | "ValueC";
```
  
- Implemented support for string indexes (maps).
  For example, the following declaration will now generate a definition for `Field` with "additionalProperties" set to a reference to `SomeType`
```
	Field : { [id : string] : SomeType };
```
  
- Added an option to allow writing the output to a file: `-o file` / `--out file`

- Added the flag `--noExtraProps` to enforce by default all objects to have `"additionalProperties": false`. The flag is set to false by default.


YousefED:
Great stuff! Thank you so much.

Could you provide tests (should be fairly easy) for the new features supported? Mainly for top-level-comments, anonymous / inline types and string indices.

otgerp:
Some more changes (+ tests):

- Changed behavior for typescript `any`: it was being translated as `{ "type": "object" }`, now it 
 generates just an empty type spec `{}` (i.e. anything will match)
  
- Changed behavior for default value `null`: it was being ignored, but now it is set as `{ "default": null }`
 
- "integer" type can now be used in a typescript file by declaring `type integer = number;`, which will be recognized as a special type (the old way of `/** @type integer **/` still works also)

- Fixed bug with creating `"title`" for enums even if `--useTitle` was set to false
  
- Fixed various comments that still were being ignored (for this I had to reorder and move quite a lot of lines, that's why so many changes). Most of the cases should work now.

- Added flag `--aliasRefs` to enforce using `$ref` for typescript alias types (e.g. "`type Vec3 = number[];`"), instead of embedding the type. This effect is also enforced now for string literal types which create enums.

- Added tests `map-types` (string indexes), `comments` (comments with annotations at different places), `type-anonymous`, and `type-primitves` (for JSON schema primitive types with default values), and enabled and fixed the test `string-literals` (the expected json was wrong).

MarcusRiemer:
This immediatly solves three problems I am facing with the current "official" version, thanks a lot to both of you for your hard work!

Issue #51, opened 2016-08-31
Is there a way to export all interfaces in a file?


21stio:
Hi,

I am looking for way to get all interfaces from one file.  So far referenced all interfaces in that file in a interface that I exported so get all the interfaces, but that's ugly and error prone..
I also tried to use '*' wildcard when calling the cli but that does not work at all.

Any idea on how to accomplish it in a nice way?

Issue #48, opened 2016-07-28
Update to typescript 2.0.


aindlq:
Typescript 2.0 has been recently released, would it make sense to upgrade the library? I tried 2.0 locally, so far so good.

YousefED:
I think this would make sense as soon as it's out of beta

aindlq:
oh sorry, I was confused. They released 2.0.0 on npm https://libraries.io/npm/typescript/2.0.0, but apparently it is still kind of beta. It seems that actually stable will be 2.0.1.

Issue #46, opened 2016-07-13
Support for intersection type


kanitw:
If I add intersection type (`&`) to my types, I get `Unsupported type` error.  

Is it possible to support this?  Thanks

Issue #44, opened 2016-06-29
Specifically allow additionalProperties on certain types?


MarcusRiemer:
I usually run the generator using `--noExtraProps true` as I don't want the API these schemas are protecting to silently discard unknown data. This works great in general, but fails in one specific cornercase: In one type I need to allow extra fields as part of a general key-value store.

In Typescript, I expressed this as

    /**
     * A widget **requires** at least a type, all other fields are
     * mandated by deriving descriptions.
     */
    export interface WidgetDescription {
        type : string
        [x: string]: any 
    }

Adding additional fields to this works fine in code:

    const foo : WidgetDescription = {
        type : "test",
        additional : 2
    }

But the generated schema does not allow any additional properties:

        "WidgetDescription": {
            "type": "object",
            "properties": {
                "type": {
                    "type": "string"
                }
            },
            "additionalProperties": false
        }

I am not quite sure how to correctly tackle this case in JSON schema. But from my understanding the presence of a property-specification like `[x: string]: any ` could translate to `additionalProperties: true`.

YousefED:
Thanks markus. Does this help?

```
/**
         * @TJS-additionalProperties true
         */
export interface WidgetDescription {
    type : string
    [x: string]: any 
}```

MarcusRiemer:
Sorry for the delay, this does solve my immediate problem, but I am still a little bit confused by the generated code. The new descriptions look like this:

    /**
     * A widget **requires** at least a type, all other fields are
     * mandated by deriving descriptions. As we don't necesarily
     * know all deriving classes at compile time (they could be
     * provided by a plugin) we poke a hole in the type system
     * here.
     * 
     * The following annotion is required to allow additional 
     * properties in the automatically generated JSON schema, see
     * https://github.com/YousefED/typescript-json-schema/issues/44
     * 
     * @TJS-additionalProperties true
     */
    export interface WidgetDescription {
        type : string
        [additional: string]: any 
    }
    
    /**
     * Describes a column that can host certain widgets
     */
    export interface ColumnDescription {
        width : number
        widgets : WidgetDescription[]
    }

The generated code places the `additionalProperties: true` property on the `ColumnDescription` hosting the `WidgetDescription`. For my use-case this is enough, but is this working as intended? I would expect the `additionalProperties: true` to pop up at the `WidgetDescription`.

    "ColumnDescription": {
        "type": "object",
        "properties": {
            "width": {
                "type": "number"
            },
            "widgets": {
[...]

PR #55, opened 2016-09-15
A few fixes and features, and TypeScript 2 support

otgerrogla:
Changes:

  - Fixed bug reported in #44 where _additionalProperties_ was being added to the property instead of the type. Checked with new test "comment-override".
    
  - Support for intersection types (issue #46). Created test "type-intersection".
  
  - Added a way to export all definitions of a file, by passing as argument `*` instead of a fully qualified type name. It generates a schema with just the "definitions" section, that includes all user-defined types parsed. (implements feature requested in #51)
  
  - Small improvement: unions of simple types like `val: number | string` now generate:
      `` "val": { "type": [ "number", "string" ] } ``
    Instead of a full "oneOf"/"anyOf" definition
    Updated test "type-union" to test both cases now, and some other tests that failed due to this change
    
  - Partial support for Typescript 2 **null**/**undefined** types (asked in #48) Added argument `strictNullChecks` (the same name used in **tsc**), which is inserted in the default CompilerOptions if provided, to enable the non-nullable-by-default behavior. Some minor changes done so that all tests pass now for both TS 1.8 and 2.0. A new test "typescript-2" is only run when 2.0 is used, and checks the null/undefined types
    
  - For compatibility (and for use in non-strictNullChecks mode), also added a @nullable annotation as requested in #36, and its test "type-nullable"


MarcusRiemer:
Wow, your `comments-override` test shows that you have exactly nailed an issue about overriding `additionalProperties` on a per-field basis that I was just about to open. Thanks a lot for your work, you have just saved me some precious time that I can now invest in my master's thesis.

YousefED:
Thanks, 

- just to make sure: "val": { "type": [ "number", "string" ] }, do you know where I can find this syntax in the JSON schema spec? Looks good, but let's make sure it corresponds to anyOf, as we just switched to anyOf as default unionModifier (probably also causing the conflict now.
- I'm looking for new maintainers, as the projects I'm working on are not actively using TJS at this moment. Let me know if you're interested

otgerrogla:
I've done another commit fixing the conflicts, and also fixed a pair of bugs with enums (and added a test "enums-mixed")

The `"val": { "type": [ "number", "string" ] }` is covered in http://json-schema.org/latest/json-schema-validation.html#anchor79 :

> 5.5.2.  type

> 5.5.2.1.  Valid values
> The value of this keyword MUST be either a string or an array. If it is an array, elements of the array MUST be strings and MUST be unique.
> String values MUST be one of the seven primitive types defined by the core specification.

> 5.5.2.2.  Conditions for successful validation
> An instance matches successfully if its primitive type is one of the types defined by keyword. Recall: "number" includes "integer"


I can have a look at the issues now and then, but I'm not sure if I'll have much free time.

YousefED:
Thanks, I found http://json-schema.org/latest/json-schema-validation.html#anchor79, but it doesn't mention how it should be interpreted.

I'll merge the request

YousefED:
closes #44 #46 #51 #48 #36

YousefED:
closes #44 
closes #46 
closes #51 
closes #48 
closes #36

low-ghost:
Huh, I was also concerned about mixed enums. Didn't know you could specify 'type' as an array and thought it'd have to be anyOf or something. And, I too was trying to solve the comments-override problem. Awesome work

PR #56, opened 2016-09-15
computed and numerical enum support

low-ghost:
This allows for computed enums (w/ binary expressions or simple operations), and numerical enums that have some initialized values and some automatically assigned values, while also fixing numerical enums, which were not quite working


YousefED:
Thanks, merged and added (made a mistake in the git workflow though)

PR #59, opened 2016-10-12
Typescript 2.0

domoritz:
This is not a complete PR but rather a starting point for a transition to ts 2.0. Please help out if you have time. 

- [ ] booleans create `anyOf`. Looks like internally booleans are `true` or `false` but there is a flag for booleans which we should be able to read. 
- [ ] Enums of strings and numbers are not handled correctly
- [ ] more?

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,13 +28,13 @@
     "schema"
   ],
   "dependencies": {
-    "typescript": "~1.8.0",
-    "glob": "~7.0.0",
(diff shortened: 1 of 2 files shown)


domoritz:
@otgerrogla would you be able to take a stab at these issues as you have worked on ts 2.0 before?

domoritz:
looks like it's been done in #61

PR #61, opened 2016-10-16
Support for typescript 2.0.3

otgerrogla:
Updated to typescript version to 2.0.3, and fixed errors due to this (mainly the enums not working because now some appear as union types (?)). All the tests pass now.

Also fixed a bug with exporting all declarations (passing * as argument), where type aliases where skipped.

Updated some tests, renamed test "typescript2" -> "strict-null-checks", and added 2 new ones: "type-union-tagged" and "type-alias-multitype-array"


otgerrogla:
@domoritz This should make it work now, but I'm not sure what you meant with the issue of the booleans creating an "anyOf". I did some changes to the handling of enums and union types, so maybe it's fixed too.

domoritz:
Sweet!

domoritz:
I just tested this pr against vega-lite and things are working well.

Issue #57, opened 2016-09-16
Extension to pr #55 and comments-override


low-ghost:
Although @MarcusRiemer's fix for comments, shown in the test for comments-override, is great for extending interfaces, it fails to account for types. Here is an example:

```typescript
/**
 * @type integer
 * @minimum 1
 */
export type PositiveInteger = number;

export interface MyObject {
  user_ids: PositiveInteger[];
  user_id: PositiveInteger;
}
```

and the expected schema is:
```json
{
    "type": "object",
    "properties": {
        "user_ids": {
            "type": "array",
            "items": {
                "$ref": "#/definitions/PositiveInteger"
            }
        },
        "user_id": {
          "$ref": "#/definitions/PositiveInteger"
        }
    },
    "required": [
        "company_ids",
        "user_id"
    ],
    "definitions": {
        "PositiveInteger": {
            "type": "integer",
            "minimum": 1
        }
    },
    "$schema": "http://json-schema.org/draft-04/schema#"
}
```
Instead, the result looks like:
```json
{
    "type": "object",
    "properties": {
        "user_ids": {
          "type": "array",
          "items": {
              "type": "number"
          }
        },
        "user_id": {
            "minimum": 1,
            "type": "integer"
        }
    },
    "required": [
        "user_ids",
        "user_id"
    ],
    "$schema": "http://json-schema.org/draft-04/schema#"
}
```
If I have a minute, I'll look into this, but a little busy. Probably isn't too complicated

domoritz:
I can confirm that this is an issue in vega-lite as well. @low-ghost, do you think you could send a pr with a fix?

domoritz:
cc @MarcusRiemer

low-ghost:
Looks like there's something to do w/ getReferencedTypeSymbol method where calling tc.getSymbolAtLocation(type.typeName) is not correct for an alias and there should also be an option of tc.getSymbolAtLocation(type.elementType.typeName), but this doesn't quite output correctly.

low-ghost:
@MarcusRiemer's fix for this is basically:
```js
this.parseCommentsIntoDefinition(reffedType, definition);
this.parseCommentsIntoDefinition(prop || symbol, returnedDefinition);
```
to
```js
this.parseCommentsIntoDefinition(reffedType, definition);
this.parseCommentsIntoDefinition(prop || symbol, symbol ? definition : returnedDefinition)
````
[...]

PR #64, opened 2016-10-20
Basic type comment functionality

low-ghost:
This pull request adds basic type comment support such that properties added to a type will be carried over into interfaces. E.g.
```js
/**
 * @type integer
 * @minimum 1
 */
export type PositiveInteger = number;

export interface MyObject {
  user_ids: PositiveInteger[];
}
```
See comments-override-types for tests.
This gets us some of the way, but test comments-override-types-tuples fails and may need a substantial rewrite. See issue [57](https://github.com/YousefED/typescript-json-schema/issues/57)


domoritz:
Please don't include changes to the .js and .js.map files in prs. It'll make it harder to merge.

low-ghost:
Sure, reverted those changes. Also, this change makes types ref-able, so it changes the functionality a bit. You might want to check the changes to test files and make sure that these changes are ok.

domoritz:
Looks good to me but it doesn't seem to work with type unions.

```ts
/**
 * An integer greater than one
 * @type integer
 * @minimum 1
 */
export type PositiveInteger = number;

foo: PositiveInteger | string;
```

Generates

```json
"foo": {
            "type": [
                "string",
                "number"
            ]
        },
```

low-ghost:
Yeah, the more I look at it, the less this change supports. Not worth adding in these changes just to have to make more substantial, less messy changes latter on. Working on some things now-ish and over the weekend perhaps

domoritz:
Thanks @low-ghost!

PR #66, opened 2016-11-01
Add test for bugged enum name generation

weswigham:
String literal unions using references to string literals nested within objects generate types with incorrect full names. This test exposes the bug. I expect line 555 of `typescript-json-schema.ts` will need to be adjusted to correctly handle this structure's type name.

@domoritz This exposes the issue I talked about [here](https://github.com/vega/vega-lite/pull/1649) - I don't have a fix yet. Do you have any insight?

### DIFF
--- a/test/programs/type-aliases-union-namespace/main.ts
+++ b/test/programs/type-aliases-union-namespace/main.ts
@@ -1,0 +1,17 @@
+
+export namespace Cardinal {
+    export const NORTH: 'north' = 'north';
+    export type NORTH = typeof NORTH;
+    export const SOUTH: 'south' = 'south';
+    export type SOUTH = typeof SOUTH;
+    export const EAST: 'east' = 'east';
+    export type EAST = typeof EAST;
+    export const WEST: 'west' = 'west';
+    export type WEST = typeof WEST;
+}
+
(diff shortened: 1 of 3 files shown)


test/programs/type-aliases-union-namespace/main.ts
```diff
@@ -0,0 +1,17 @@
...
+    export type NORTH = typeof NORTH;
+    export const SOUTH: 'south' = 'south';
+    export type SOUTH = typeof SOUTH;
+    export const EAST: 'east' = 'east';
+    export type EAST = typeof EAST;
+    export const WEST: 'west' = 'west';
+    export type WEST = typeof WEST;
+}
+
+export type Cardinal = Cardinal.NORTH | Cardinal.SOUTH | Cardinal.EAST | Cardinal.WEST;
```
domoritz: Make this `export type Cardinal = typeof Cardinal.NORTH` and remove `export type NORTH = typeof NORTH;` above.

domoritz:
I don't know what's causing the issue yet but with the test it should be easy to fix. Do you want to take a stab at it?

weswigham:
I likely won't have a chance to take a stab at it for a day or two. If you could manage it, that'd be great.

domoritz:
Now in #68

PR #76, opened 2017-01-11
Typescript 2.1

domoritz:


domoritz:
The problem with the missing docs is https://github.com/Microsoft/TypeScript/issues/11956

domoritz:
@YousefED works in TS 2.1 :-D

PR #78, opened 2017-01-16
Fixes for aliased types

domoritz:


domoritz:
@mrix There are still two tests failing (commented out). If you get the chance, could you try to find a fix? I still think the changes in this PR are improvements.

PR #81, opened 2017-01-24
Added buildGenerator function to expose the loaded JsonSchemaGenerator

zakhenry:
This is needed in my use case as I want to parse all files once then ad-hoc build json schema's based on another source.


domoritz:
Thanks for the pull request. Can you make sure that test pass, add a test, and not include the built files (js and js.map)?

zakhenry:
@domoritz I've added a test and also cleaned up a couple of fixtures that had lint errors on Travis. Please take another look

PR #86, opened 2017-01-30
fix for issue 85 - Generates wrong schema for readonly arrays

mmc41:
Please review and if accepted let me know when a new npm can be published as well?


domoritz:
Merged with slight modifications and released as 0.8.

Issue #92, opened 2017-02-01
type alias lose annotations


nidi3:
````
/**
 * @format uuid
 */
export type UUID=string

interface A{
    u:UUID
}
````

gives

````
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "definitions": {
        "A": {
            "properties": {
                "u": {
                    "format": "uuid",
                    "type": "string"
                }
            },
            "type": "object"
        },
        "UUID": {
            "type": "string"
        }
    }
}

```` 

I'd expect `format: uuid` in the UUID definition.
This prevents importing annotated type aliases, which would be a really cool thing.

**Comment by domoritz** [comment]
This is odd. Can you send a pull request with this test case so we can look into it? Maybe you can even take a stab at a fix. The code for this tool is actually not that large.

**Comment by bdarcus** [comment]
I seem to see similar; format doesn't get added to the schema.

```typescript
/**
 * An EDTF level 0 or 1 date, with optional seasonal range.
 *
 * See https://github.com/retorquere/json-schema-edtf
 *
 * @format edtf/level-1+season-intervals
 */
export type EDTFDATE = string | null | undefined;
```

Flynamic:
aNy UpDaTeS oN tHiS

Flynamic:
Any updates on this?

PR #93, opened 2017-02-02
add test case for not directly used annotated type aliases

nidi3:
see #92 
I also tried to fix it, but I fear it will cause quite some changes.
As I see it, `allSymbols` contains `Type`s. These are resolved types (string instead of MyString in the test case) and the annotation information is lost.
So `allSymbols` should probably contain `Symbol`s or `Node`s to preserve the unresolved types.

### DIFF
--- a/test/programs/type-aliases-annotated-import/def.ts
+++ b/test/programs/type-aliases-annotated-import/def.ts
@@ -1,0 +1,5 @@
+/**
+ * My string
+ * @format pattern [a-z]
+ */
+export type MyString = string;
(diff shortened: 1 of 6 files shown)


domoritz:
Yes, you are right. I think we need to make these changes and use symbols instead of types.

domoritz:
I ended up modifying a different library to get correct aliases: https://github.com/vega/typescript-to-json-schema. It uses nodes instead of types but I don't see why symbols wouldn't work. https://astexplorer.net/ was super helpful.

domoritz:
I don't think we can really fix this issue with the current approach.

Issue #100, opened 2017-02-25
@default be an array


navossoc:
Hello, I would like to know if it is possible to set a default value as an array, something like this:

**Test case:**
```ts
class Test {
  /**
   * @default ["a", "b", "c"]
   */
  myKey: string[];
}
```

**Current output**
```json
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "properties": {
        "myKey": {
            "default": "[\"a\", \"b\", \"c]",
            "items": {
                "type": "string"
            },
            "type": "array"
        }
    },
    "type": "object"
}
```

**Expected output:**
```json
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "properties": {
        "myKey": {
            "default": ["a", "b", "c"],
            "items": {
                "type": "string"
            },
            "type": "array"
        }
    },
    "type": "object"
}
```

According to the [docs](http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.2):

> 6. Metadata keywords
> 6.2. "default"
> 
> There are no restrictions placed on the value of this keyword.

Thanks for the help...

domoritz:
Yes, this is a bug.

navossoc:
For sure, I can give it a try...

I'm new at this project and I have a exam on the next week.
So after that I will reserve some time to take a look and find a proper way to implement it.

Is there any contribution guideline to follow or something like that?

[]'s

navossoc:
Well, I sent a PR.  That was the best way I could think to fix this in an easy way.
Let me know what you think...

PR #101, opened 2017-02-26
default annotation should accept any value

navossoc:
This PR solves the problem mentioned on the issue #100.

There is a minor issue noted, for now it can't accept a object as a default value.

`@default {key: "value"}` fails

A quick workaround is to use quotes around the key name, so it will be treated like a valid JSON and correctly parsed as an object.

`@default {"key": "value"}` works

I think it can be improved later, but for now this PR adds a new feature without breaking the existing ones.

### DIFF
--- a/test/programs/annotation-default/main.ts
+++ b/test/programs/annotation-default/main.ts
@@ -1,0 +1,38 @@
+interface MyObject {
+  /**
+   * @default true
+   */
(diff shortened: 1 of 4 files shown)


domoritz:
Sweet. Thanks!

domoritz:
Released as v0.9.2

PR #102, opened 2017-02-27
feat: typeOf keyword for functions

develar:
Custom JSON-Schema keywords for Ajv validator — https://github.com/epoberezkin/ajv-keywords#typeof

Problem: no way to validate that option is function (https://github.com/electron-userland/electron-builder/wiki/Options#Config-afterPack).


domoritz:
Thank you for the PR. Just to clarify, you only implemented `typeOf` for functions but not other types. Is there a particular reason?

develar:
> Is there a particular reason?

No need. JSON Schema doesn't support only function validation. `symbol` is covered by enum and not related.

domoritz:
Can you make this more explicit and describe it in the command doc?

develar:
Added to readme. Changed to 

> Use `typeOf` keyword (https://goo.gl/DC6sni) for functions.

PR #104, opened 2017-03-10
add-keywords: option for user provided validationKeywords

low-ghost:
This allows passing in specific keywords that the user might want in addition to the standard validation keywords. Specific use case: faker or chance in combination with json-schema-faker. Allowing custom keywords will be useful in other scenarios as well (I like errorMessage to help when validation of a schema fails). Added tests and yargs config.


domoritz:
This is great. Can you update the readme and also remove the compiled files from this pr as it makes it harder to merge your changes.

low-ghost:
Yep, just thought about the readme additions

Issue #95, opened 2017-02-06
Incorrect interaction with TypeScript "Mapped types" feature


benny-medflyt:
TypeScript has a new feature (2.1) called "Mapped Types".

The example from the release notes:

```typescript
interface Person {
    name: string;
    age: number;
    location: string;
}
```

If we want a type that looks like this:

```typescript
interface PartialPerson {
    name?: string;
    age?: number;
    location?: string;
}
```

then instead of writing out all the fields, you can use:

```typescript
type PartialPerson2 = Partial<Person>;
```

When typescript-json-schema is run on the manually written `PartialPerson` type, it works great, but when run on the "magically" created `PartialPerson2` it produces the wrong JSON schema. (`PartialPerson` and `PartialPerson2` should be equivalent)

domoritz:
These things are tricky because we need special cases in the code. I don't think the Typescript compiler has a way to resolve generics before we get the types.

benny-medflyt:
Thanks for the explanation. I don't know about the internals of the TypeScript compiler, but it looks like at least `tsserver` is able to get detailed information about the type:

<img width="217" alt="screen" src="https://cloud.githubusercontent.com/assets/24827547/22705929/160614ce-ed76-11e6-9acc-aace458e42bf.png">

domoritz:
Interesting. It may be a good idea to use the tsserver then instead of calling compiler internal functions.

domoritz:
The server calls `getTypeDefinitionAtPosition` https://github.com/Microsoft/TypeScript/blob/5644b013c999709be0b75508d829449f058f17c1/src/server/session.ts#L470

TimMensch:
I just ran into this right now. FWIW you don't need to use generics to hit this bug:

```
declare interface DiceSpec {
    basic : number;
    air : number;
    fire : number;
    water : number;
    earth : number;
}

type MaybeDiceSpec = {
    [ P in keyof DiceSpec ]?: DiceSpec[ P ];
};
```

The MaybeDiceSpec comes out as an empty object. Just like above, tsserver can look at the type and see exactly what the derived type looks like.

domoritz:
@TimMensch Yeah, keyof is a very new feature that is not supported yet. Again, I think if we can somehow use the language server, we might be able to not have to manually adjust the library every single time ts adds a new feature.

HoldYourWaffle:
[...]

PR #105, opened 2017-03-17
Handling mapped type Partial

benbabic:
Guys, 

This PR provides a test case for ts type `Partial` and an implementation that passes the test without breaking the existing ones.
I did not update the README.md as I did not know what to put in there but I'm more than happy to do it if you want.

Not really sure it works for all mapped types but it works for `Partial`

Please read the commit messages for technical details.

Relates to #95

### DIFF
--- a/test/programs/partial-mapped-type/main.ts
+++ b/test/programs/partial-mapped-type/main.ts
@@ -1,0 +1,4 @@
+type MyObject = Partial<{
+    required: number;
+    optional: number;
+}>
--- a/test/programs/partial-mapped-type/main.ts
+++ b/test/programs/partial-mapped-type/main.ts
@@ -1,4 +1,4 @@
 type MyObject = Partial<{
     required: number;
     optional: number;
-}>
+}>;
--- a/test/programs/partial-mapped-type/schema.json
+++ b/test/programs/partial-mapped-type/schema.json
@@ -1,0 +1,12 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "type": "object",
+    "properties": {
+        "required": {
+            "type": "number"
+        },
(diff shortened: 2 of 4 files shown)


test/programs/partial-mapped-type/main.ts
```diff
@@ -0,0 +1,4 @@
+type MyObject = Partial<{
+    required: number;
+    optional: number;
```
domoritz: It's a bit confusing that both variables are optional but one has the name `required`. Why not only one?

benbabic: Indeed, I just lazily copy / pasted the optional test. Will change that to something more meaningful later today. Any suggestions for the names ?

domoritz:
Can you create a nested object with one object that is partial and one that
is not? I'd like to test that partial actually works and not also pass if
required is disabled.

On Sat, Mar 18, 2017, 13:24 Benjamin Babic <notifications@github.com> wrote:

*@benbabic* commented on this pull request.
------------------------------

In test/programs/partial-mapped-type/main.ts
<https://github.com/YousefED/typescript-json-schema/pull/105#discussion_r106786196>
:

> @@ -0,0 +1,4 @@
+type MyObject = Partial<{
+    required: number;
+    optional: number;

Indeed, I just lazily copy / pasted the optional test. Will change that to
something more meaningful later today. Any suggestions for the names ?

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/YousefED/typescript-json-schema/pull/105#discussion_r106786196>,
or mute the thread

domoritz:
Any new on this @benbabic?

domoritz:
Closing because of inactivity.

PR #118, opened 2017-06-01
Fixed types with same name in other files being overwritten

otgerrogla:
Fixes problem with "imports" where if a type was declared in the main file, and one existed in another file with the same name, only the first would be created in the schema (see the test).


domoritz:
I currently have a large number of local changes to resolve conflicts like this. I'm trying to get rid of `getSymbolAtLocation`. Let me see how far I get before merging this. 

Thanks for the PR!

Issue #122, opened 2017-06-05
Interfaces with properties typed as a specific enum value throw an error


marcindobry:
When defining an interface with one of the properties being tied to a specific value of an enum instead of an enum type, an error is thrown. The code assumes it only uses EnumDeclaration but this is an EnumMember.
```ts
export enum HTTPStatusCode {
  Ok = 200,
  BadRequest = 400,
  InternalServerError = 500,
};

export interface BrokenSpecificResponse {
  code: HTTPStatusCode.Ok;
};

// This is what it logically means
export interface ExpectedSpecificResponse {
  code: 200;
};
```

```
node typescript-json-schema.js example.ts BrokenSpecificResponse
<repo-path>/typescript-json-schema/typescript-json-schema.js:270
        enm.members.forEach(function (member) {
                   ^

TypeError: Cannot read property 'forEach' of undefined
    at JsonSchemaGenerator.getEnumDefinition (<repo-path>/typescript-json-schema/typescript-json-schema.js:270:20)
    at JsonSchemaGenerator.getTypeDefinition (<repo-path>/typescript-json-schema/typescript-json-schema.js:590:26)
    at JsonSchemaGenerator.getDefinitionForProperty (<repo-path>/typescript-json-schema/typescript-json-schema.js:225:31)
    at <repo-path>/typescript-json-schema/typescript-json-schema.js:436:37
    at Array.reduce (native)
    at JsonSchemaGenerator.getClassDefinition (<repo-path>/typescript-json-schema/typescript-json-schema.js:434:45)
    at JsonSchemaGenerator.getTypeDefinition (<repo-path>/typescript-json-schema/typescript-json-schema.js:597:26)
    at JsonSchemaGenerator.getSchemaForSymbol (<repo-path>/typescript-json-schema/typescript-json-schema.js:614:24)
    at generateSchema (<repo-path>/typescript-json-schema/typescript-json-schema.js:742:32)
    at exec (<repo-path>/typescript-json-schema/typescript-json-schema.js:772:22)
```

marcindobry:
Closing since the solution got merged.

PR #123, opened 2017-06-05
Fix using enum values in interfaces

marcindobry:
Fix for https://github.com/YousefED/typescript-json-schema/issues/122
***
Please:

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -361,7 +361,15 @@ export class JsonSchemaGenerator {
     private getEnumDefinition(clazzType: ts.Type, tc: ts.TypeChecker, definition: Definition): Definition {
         const node = clazzType.getSymbol().getDeclarations()[0];
         const fullName = tc.typeToString(clazzType, undefined, ts.TypeFormatFlags.UseFullyQualifiedType);
-        const enm = <ts.EnumDeclaration>node;
+        let enm: ts.EnumDeclaration | ts.EnumMember;
+        let members: ts.EnumMember[];
+        if (node.kind === ts.SyntaxKind.EnumMember) {
+            enm = <ts.EnumMember>node;
+            members = [enm];
+        } else {
+            enm = <ts.EnumDeclaration>node;
+            members = enm.members;
+        }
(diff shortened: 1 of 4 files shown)


typescript-json-schema.ts
```diff
@@ -372,7 +380,7 @@ export class JsonSchemaGenerator {
             }
         };
 
-        enm.members.forEach(member => {
+        members.forEach(member => {
```
domoritz: why not just `(node.kind === ts.SyntaxKind.EnumMember ? [enm] : enm.members).forEach(...)`?

typescript-json-schema.ts
```diff
@@ -361,7 +361,15 @@ export class JsonSchemaGenerator {
     private getEnumDefinition(clazzType: ts.Type, tc: ts.TypeChecker, definition: Definition): Definition {
         const node = clazzType.getSymbol().getDeclarations()[0];
         const fullName = tc.typeToString(clazzType, undefined, ts.TypeFormatFlags.UseFullyQualifiedType);
-        const enm = <ts.EnumDeclaration>node;
+        let enm: ts.EnumDeclaration | ts.EnumMember;
```
domoritz: I think this can be const

domoritz:
Thanks a lot for the test case and PR. Can you fix it so that the test pass?

typescript-json-schema.ts
domoritz: See how I implemented it here: https://github.com/vega/typescript-to-json-schema/commit/462fe6d6a9ef7a52cd56edb8cfe930dd9f466a54. 

Btw, thanks for the PR. It helped me to also fix a bug in the other library I'm working on.

marcindobry:
@domoritz Made the assignment more compact. 

The failing test is actually on master due to this change https://github.com/YousefED/typescript-json-schema/pull/119. I can make a separate PR to fix that, but I'm just unsure whether the tests are failing for a correct reason or just a mistake in defining the expected schema.

domoritz:
Fantastic. Thanks!

Issue #120, opened 2017-06-03
Allowing non-standard names for `tsconfig.json`


MarcusRiemer:
I have an Angular CLI project that makes use of non-standard names for the `tsconfig.json` file. The concrete name is `tsconfig.app.json`, which results in the following error message:

    File 'scratch-sql/client/src/tsconfig.app.json' has unsupported extension. The only supported extensions are '.ts', '.tsx', '.d.ts'.

Would it be possible to soften the rules for names of project files?

domoritz:
Can you send a pull request?

PR #124, opened 2017-06-05
Allowing non-standard names for `tsconfig.json`

MarcusRiemer:
This simply replaces the previous strict check (`name === "tsconfig.json"`) with a more relaxed check that simply ensures that the name begins with `tsconfig` and ends in `.json`. This is required in the case of e.g. Angular applications that use names like `tsconfig.app.json` and `tsconfig.spec.json` (see #120).

I have no idea how to properly test this, as the current testing infrastructure only seems to work directly with `.ts` files. The regex still matches `tsconfig.json` though, and I would not expect it to break any existing configurations as `.json` was never a allowed ending for source files.

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -7,7 +7,7 @@ import * as stringify from "json-stable-stringify";
 const vm = require("vm");
 
 const REGEX_FILE_NAME = /".*"\./;
-const REGEX_TSCONFIG_NAME = /^tsconfig.*\.json$/;
+const REGEX_TSCONFIG_NAME = /^.*\.json$/;
 const REGEX_TJS_JSDOC = /^-([\w]+)\s([\w-]+)/g;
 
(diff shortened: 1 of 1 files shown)


domoritz:
Why not just a `.json`?

MarcusRiemer:
The file, [according to the standard](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html), is named `tsconfig.json` after all. I wanted to deviate from this as little as possible. The proper way would probably be to actually inspect the `.json` file at point. [Thankfully there is a spec available](http://json.schemastore.org/tsconfig), but this seems out of scope for this tool.

In the end you can call the shots. If you are sure enough that the `.json` ending will never, ever be used for anything else then the `tsconfig`, widening this seems reasonable.

domoritz:
Yeah, let's just support files ending in `.json`. Thanks for the references.

domoritz:
Thank you!

PR #125, opened 2017-06-07
[Fixes #106] Don't add ignored properties to "required" arrays

### LINKED ISSUE #106: Properties marked with @ignore are still listed in required array
When I mark properties with `@ignore`, I expect them to be removed from the schema altogether. But these properties still show up when I run `typescript-json-schema --required`.

Minimal test case:

```typescript
// test.ts
class Test {
    /**
     * @ignore
     */
    ignoreMe: string;
}
```

Now, when I run `typescript-json-schema tsconfig.json Test`, I'd expect this output:

```json
{
    "properties": {
    },
    "type": "object"
}
```

But I get this instead:

```json
{
    "properties": {
    },
    "required": [
        "ignoreMe"
    ],
    "type": "object"
}
```

I don't think this is useful behavior, because it doesn't make sense to require a property but not mention it elsewhere in the schema.

This is using `typescript-json-schema@0.10.0` with `typescript@2.1.6`.

RyanLB:
This PR stops ignored properties from being added to the "required" arrays in generated schemas.

We do this by replacing the conditional 
```
if (!(prop.flags & ts.SymbolFlags.Optional) && !(<any>prop).mayBeUndefined) {
    required.push(prop.getName());
}
```
with the following, stricter one:
```
if (!(prop.flags & ts.SymbolFlags.Optional)
    && !(<any>prop).mayBeUndefined
    && prop['tags'].find((x: any) : boolean => { 
    return x.name === 'ignore';
}) === undefined) {
    required.push(prop.getName());
}
```

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -584,7 +584,11 @@ export class JsonSchemaGenerator {
             }
             if (this.args.generateRequired) {
                 const requiredProps = props.reduce((required: string[], prop: ts.Symbol) => {
-                    if (!(prop.flags & ts.SymbolFlags.Optional) && !(<any>prop).mayBeUndefined) {
+                    if (!(prop.flags & ts.SymbolFlags.Optional)
+                        && !(<any>prop).mayBeUndefined
+                        && prop["tags"].find((x: any): boolean => {
+                            return x.name === "ignore";
+                        }) === undefined) {
                         required.push(prop.getName());
                     }
                     return required;
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -584,11 +584,9 @@ export class JsonSchemaGenerator {
             }
             if (this.args.generateRequired) {
                 const requiredProps = props.reduce((required: string[], prop: ts.Symbol) => {
-                    if (!(prop.flags & ts.SymbolFlags.Optional)
-                        && !(<any>prop).mayBeUndefined
-                        && prop["tags"].find((x: any): boolean => {
-                            return x.name === "ignore";
-                        }) === undefined) {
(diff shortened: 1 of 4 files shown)


typescript-json-schema.ts
```diff
@@ -584,7 +584,11 @@ export class JsonSchemaGenerator {
             }
             if (this.args.generateRequired) {
                 const requiredProps = props.reduce((required: string[], prop: ts.Symbol) => {
-                    if (!(prop.flags & ts.SymbolFlags.Optional) && !(<any>prop).mayBeUndefined) {
+                    if (!(prop.flags & ts.SymbolFlags.Optional)
+                        && !(<any>prop).mayBeUndefined
+                        && prop["tags"].find((x: any): boolean => {
```
domoritz: Can you reuse `otherAnnotations` or just call `parseCommentsIntoDefinition` instead of reading the annotations again?

RyanLB: Sure, how's that look?

Also, let me know when you're ready to merge and I'll squash my commits.

domoritz:
Thank you for the PR!

domoritz:
Hmm, looking at this PR again, @RyanLB, why does `ignored` not appear in the output at all? I thought ignore should only not put properties in the required list.

RyanLB:
The `@ignore` attribute always stopped member from appearing in the `properties` map. Was that not intentional? I'd been regarding this as a bug because properties that didn't otherwise exist in the schema were showing up in the `required` array.

domoritz:
Right, makes sense. I just added a @hide annotation to https://github.com/vega/typescript-to-json-schema with the same behavior.

PR #126, opened 2017-06-15
Added test to ensure correct behaviour for derived optionals

MarcusRiemer:
I am currently facing some problems with incorrectly generated required fields in my project. This was an attempt to create a minimal failing test, but "sadly" this new test doesn't actually fail. But as I have written the new test you may want to integrate it anyway.


domoritz:
Can you try https://github.com/vega/typescript-to-json-schema? Note that my fork is NOT released on npm.

Issue #131, opened 2017-06-27
Support external references


jakefurler:
I want to reference a schema that's hosted externally, from within my schema. This is done in json schema using $ref, but there's currently no way to control that manually with this library afaik.

PR #132, opened 2017-06-27
Issue #131 add support for $ref annotation

jakefurler:
I want to be able to specify an external reference in my schema. Doing it with an annotation like this seems like the simplest way, but it does override any value for $ref that it would otherwise have, as you can see in the test. Please let me know if this is acceptable or if there's a better approach.

### DIFF
--- a/test/schema.test.ts
+++ b/test/schema.test.ts
@@ -175,6 +175,8 @@ describe("schema", () => {
      */
 
     assertSchema("annotation-default", "main.ts", "MyObject");
+    assertSchema("annotation-ref", "main.ts", "MyObject");
+    assertSchema("annotation-tjs", "main.ts", "MyObject");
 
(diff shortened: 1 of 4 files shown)


test/schema.test.ts
```diff
@@ -175,6 +175,8 @@ describe("schema", () => {
      */
 
     assertSchema("annotation-default", "main.ts", "MyObject");
+    assertSchema("annotation-ref", "main.ts", "MyObject");
+    assertSchema("annotation-tjs", "main.ts", "MyObject");
```
jakefurler: "annotation-tjs" is unrelated, I just noticed the tests weren't being run.

jakefurler:
Looks like the test failure is a timeout, shouldn't be related to these changes.

domoritz:
I restarted the failing job. Not sure why we get these timeouts.

jakefurler:
Thanks for the merge @domoritz! Any chance of a release so I can use these changes?

PR #149, opened 2017-10-06
Add setting to ignore private members. Fixes #142

### LINKED ISSUE #142: Exclude private variables?
Hi, can the generator exclude private variables?

Example
```js
class Animal {
    fur: boolean
    private db: any
}
```

Should not include db in schema.json

lev-kuznetsov:
Please:
- [X] Do not include the compiled `.js`, `.js.map`, or `.d.ts` in your pull request as it makes it harder to merge your changes. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ Options:
   --out, -o             The output file, defaults to using stdout
   --validationKeywords  Provide additional validation keywords to include            [array]   [default: []]
   --ignoreErrors        Generate even if the program has errors.                     [boolean] [default: false]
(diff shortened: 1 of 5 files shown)


domoritz:
Can you add a test for interfaces?

lev-kuznetsov:
Do you mean something like this:

```
interface Foo {
  private bar: string
}
```

You can't use private in an interface - that's a compile error, at least for me. Or is it something else?

domoritz:
> You can't use private in an interface - that's a compile error, at least for me. Or is it something else?

Never mind, I wasn't thinking.

PR #150, opened 2017-10-09
Added option to use fully qualified names.

### LINKED ISSUE #112: conflicting type names from different files
Hello,

We just ran into this "[TODO](https://github.com/YousefED/typescript-json-schema/blob/cf275f3e7c4280ab0b67aa985ea44a0f52d54c19/typescript-json-schema.ts#L825)" (and of course into the typescript version upgrade issue too, we used `object` and that was not recognized).

What'd be the best way to fix the conflicting names? For a quick workaround removing that replace helps. Then symbol names become `"/home/some/thing/path/to/the/module".Whatever`, without the `.ts` at the end of the file path.

fabiandev:
This is an approach to handle duplicate type names via the API, while preserving backward compatibility through the newly introduced option `fullyQualifiedNames` which defaults to `false`. Fixes #112.


fabiandev:
Will open a new pull requests with less commits.

PR #151, opened 2017-10-09
Add uniqueNames option.

### LINKED ISSUE #112: conflicting type names from different files

fabiandev:
This is an approach to handle duplicate type names via the API, while preserving backward compatibility through the newly introduced option `uniqueNames` which defaults to false. Fixes #112.


domoritz:
Can you add a test?

domoritz:
I don't understand the test. It just tests whether string constants work, doesn't it?

fabiandev:
It tests whether different types of the same name can be obtained with the `uniqueNames` option. I've just added the string literal to easily see that both types `MyObject` from `main.ts` and `other.ts` can be generated. Since there are other tests that actually test the JSON schemas, this one should just verify that the newly introduced option `uniqueNames` works alongside the `getSymbols` method.

domoritz:
Thank you! Shouldn't this option be the default? Could you rebase against the latest master?

fabiandev:
Yes, probably this option could be default, but I wanted this pull request to be backward compatible. I will revisit this PR and will also resolve the conflicts.

domoritz:
Thank you @fabiandev

domoritz:
Could you rebase the branch so that I can merge it?

typescript-json-schema.ts
```diff
@@ -854,6 +866,16 @@ export class JsonSchemaGenerator {
         return root;
     }
 
+    public getSymbols(): SymbolRef[];
```
domoritz: What does this method do here?

domoritz: What does this method do here?

fabiandev: Thought it may be more clear to use function overloading here. Could also be removed, since the implementation with the optional parameter should do.

domoritz:
Thank you!

PR #155, opened 2017-10-19
Added excludeFunctions option

KurtGokhan:
Created 2 tests. One for the most basic functionality of ignoring function. Other test is for union types.
Edited Readme to include 'excludeFunctions' option. 

CLI arguments needs to be updated and the actual functionality must be implemented.


domoritz:
We are ignoring functions now.

PR #157, opened 2017-10-25
Jsdoc parsing for raw types

azaslavsky:
The test should make it pretty clear what this fixes, but the broken behavior was that this type:
```typescript
/**
 * This is a description
 * @pattern ^mystring-[a-zA-Z0-9]+$
 * @minLength 10
 * @maxLength 24
**/
type MyString = string;
```

would produce this unannotated JSON Schema:

```json
{
    "type": "string"
}
```

This has now been rectified, though a small breaking change was made to the API of `JsonSchemaGenerator` in the process.  The property `userSymbols` now contains actual `Symbols`, rather than their `Type` definitions.  Since `allSymbols` still contains the`Type` definitions, and `userSymbols` is a subset of it, it is still possible to get a type for a `userSymbol` by doing something like:

```typescript
//userTypes would look exactly like the old userSymbols object
const userTypes: { [name: string]: ts.Type } = {};
for (let i in userSymbols) {
  userTypes[i] = allSymbols[i];
}
```

I do feel like `allSymbols` should either be renamed (`allTypes`) or should contain `Symbols` like `userSymbols` going forward, but I think that would need to be a separate PR.  Thoughts?


domoritz:
Thank you! I think we used to use Symbols instead of types, which is why the variable is called `allSymbols`. 

The fix looks good to me. I also tested the example above in https://github.com/vega/ts-json-schema-generator and it worked without problems. 

```
npm run run -- --path test.ts --type MyString                                                                                                                                                                                                                        4432ms

> ts-json-schema-generator@0.14.0 run /Users/domoritz/Developer/UW/ts-json-schema-generator
> ts-node ts-json-schema-generator.ts "--path" "test.ts" "--type" "MyString"

{
  "$ref": "#/definitions/MyString",
  "definitions": {
    "MyString": {
      "description": "This is a description",
      "maxLength": 24,
      "minLength": 10,
      "pattern": "^mystring-[a-zA-Z0-9]+$",
      "type": "string"
    }
  }
}
```

azaslavsky:
Hmm, strange, I tried using ts-json-schema-generator as a drop-in replacement to fix this problem, but got the same output.  Maybe there's something else that effects it?

domoritz:
Hmm, as you can see, I ran `npm run run -- --path test.ts --type MyString` with

```ts
/**
 * This is a description
 * @pattern ^mystring-[a-zA-Z0-9]+$
 * @minLength 10
 * @maxLength 24
 */
export type MyString = string;
```

I'm using `v0.14`. Sidenote: `ts-json-schema-generator` is much more opinionated and optimized for my use case in https://github.com/vega/vega-lite.

azaslavsky:
It's definitely likely that I pulled the wrong version, or otherwise screwed something up on my end.

The more restrictive API makes it difficult for me to switch (ex: I use `validationKeywords` here a good amount).  Also, the new README in that repo doesn't have a whole lot of documentation, so I wasn't quite sure how to implement it properly.  Either way, I'm quite happy with this library, and am happy to fix any bugs I stumble upon while using it! 💯

domoritz:
"If it works, don't fix" it or "If the library you use works for you, don't switch". I had issues with aliases that I couldn't address with the approach of using types/symbols so I built a different library that operates on the AST.

PR #158, opened 2017-10-26
Fix regex bug for "@TJS-pattern"

azaslavsky:
Currently, this causes a fatal error:
```typescript
export interface Foobar {
  /**
   * This foo describes my bar
   * @TJS-pattern ^[somefunregex]+$
   */
  foo: string;
}
```

This is due to the regex for `REGEX_TJS_JSDOC` being too restrictive.  While I realize that in the case of `@TJS-pattern` a simple `@pattern` could easily be substituted, it seems like the `TJS-` prefix should maintain consistent behavior across all JSON Schema tags.

### DIFF
--- a/test/programs/annotation-tjs/main.ts
+++ b/test/programs/annotation-tjs/main.ts
@@ -49,4 +49,14 @@ interface MyObject {
      * @TJS-pattern ^[a-zA-Z0-9]{4}-abc_123$
      */
     regexPattern: string;
+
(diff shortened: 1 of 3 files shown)


domoritz:
Thanks! This is not an issue in ts-json-schema-generator because I don't support arbitrary `TJS-` annotations. That's the advantage of a more focused library ;-)

PR #160, opened 2017-11-01
add .js.map file to .npmignore

nchen63:
since the .ts isn't published, you don't need the map file


domoritz:
Doesn't the map file provide a view into the source?

PR #161, opened 2017-11-01
Fix empty string literal error

### LINKED ISSUE #159: Fails with empty string literal
With the input:
```
export type MeasurementSuffix = "calc" | "" | "%" | "px";

export interface IBoop {
    x: MeasurementSuffix;
}
```

calling `typescript-json-schema file.ts IBoop` results in `Unsupported type:  `. Taking out the empty string fixes the issue.

nchen63:
Please:

fixes #159


nchen63:
the `parseFloat(str as string);` is hacky, but that's essentially what it was doing before. I tested `parseFloat(1)` in Chrome and it works at least.

domoritz:
Thank you.

PR #166, opened 2017-11-09
Add typescript type void, fix invalid 'undefined' type translation

zyrorl:
Undefined is not a valid type under draft-04 json schema, removed it to function the same way as any, and likewise with void TS type.

Please:
- [] Provide a test case & update the documentation in the `Readme.md`

### DIFF
--- a/test/programs/strict-null-checks/main.ts
+++ b/test/programs/strict-null-checks/main.ts
@@ -2,7 +2,8 @@
 class MyObject {
      val: number;
      valNullable: number | null;
-     valUndef: number | undefined;
+     valUndef: undefined;
+     valVoid: void;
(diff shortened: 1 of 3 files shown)


domoritz:
Tests are failing.

zyrorl:
@domoritz - im fixing now

zyrorl:
This should now be fixed @domoritz

domoritz:
Hmm, the schema in the changed tests doesn't look right to me. Can you explain why each of the changes is right?

zyrorl:
We were trying to get rid of undefined as a type in json schema because it's invalid, but this is the wrong way to go about i8t, ill close the PR

PR #170, opened 2017-11-23
Add includeReffedDefinitions flag for getSchemaForSymbols

### LINKED ISSUE #169: getSchemaForSymbols not returning referenced symbols
When I call `getSchemaForSymbol` for a symbol with referenced symbols they also are included in the generated output. When I do the same for `getSchemaForSymbols` they aren't.

I could call `getSchemaForSymbol` for each symbol, but since I have a lot of shared definitions then I would need to merge these, and sometimes I get namespace collisions.

It would be nice if `getSchemaForSymbols` returned referenced symbols like `getSchemaForSymbol` does.

aboyton:
Fixes #169

Please:

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -848,7 +848,7 @@ export class JsonSchemaGenerator {
         return def;
     }
 
-    public getSchemaForSymbols(symbolNames: string[]): Definition {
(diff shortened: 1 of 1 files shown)


domoritz:
Good catch. Thank you!

aboyton:
Thanks for the quick release and version patch. This is working great for me now.

PR #180, opened 2018-01-21
Convert compiler options from JSON format to Typescript format, adding an optional argument for base path

peterjwest:
The base path was added to provide flexibility for programmatic users, however I could remove it.
I had a look at tests and couldn't see any obvious place to add one. Any suggestions?

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,10 @@ const compilerOptions: TJS.CompilerOptions = {
     strictNullChecks: true
 }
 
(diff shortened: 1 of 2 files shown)


domoritz:
Please change the ts code, not the js!

peterjwest:
Oops, sorry.

peterjwest:
Updated! Annoyingly the input type for `convertCompilerOptionsFromJson` is `any`, but this may be a necessary consequence of accepting any old JSON data. It returns `ts.CompilerOptions` though.

PR #187, opened 2018-04-01
Move CLI functionality to separate module to make library usage cleaner

dvdsgl:
I'm deploying `typescript-json-schema` as a dependency within a web app, but my deploy fails because `yargs` cannot be minified (because it's es6). `yargs` is only used when  `typescript-json-schema` is used as a CLI tool, so I've moved the `run()` function and CLI functionality to `typescript-json-schema-cli.ts`.

This allows consumers of `typescript-json-schema` as a library to ignore `yargs` and any other CLI/node dependencies in the future.

Please:


dvdsgl:
I found your requirement "Do not include the compiled .js, .js.map, or .d.ts in your pull request..." confusing since you have those files for your main module committed to the repo. Could you please remove them to make this clearer?

domoritz:
Thank you for the PR. Can you update the readme accordingly?

I don't want to make the call for this alone. @otgerrogla, what are your thoughts?

dvdsgl:
I've updated the README.

dvdsgl:
Your PR instructions asked me not to include `.js`, `.js.map`, or `.d.ts` files, so the `typescript-json-schema.js` file in this branch still reflects the contents of the previous `typescript-json-schema.ts` file.

How should we resolve this? Do you want to include the compiled version of the code or not?

domoritz:
I will build them myself. The reason why we don't want them included is that it would otherwise cause merge conflicts.

dvdsgl:
FYI: I just published [the app](https://app.quicktype.io/?s=Confetti.ts&l=cs&features=just-types) with this feature.

domoritz:
That's super awesome! I've seen someone use quicktype before to generate Rust code for [Vega-Lite](https://vega.github.io/vega-lite/).

Btw, I am also the developer of https://github.com/vega/ts-json-schema-generator, which we switched to for Vega-Lite as it is a bit more robust (but also opinionated and doesn't have as many configuration options).

domoritz:
Ahh, here is the tweet: https://twitter.com/ArmyOfBruce/status/965944090247352320

domoritz:
Since this doesn't break anything, I'm going ahead and merge.

dvdsgl:
Oh, awesome, I will try the other project.

PR #188, opened 2018-04-01
Remove build artifacts

dvdsgl:
Your PR guidelines ask:

  > Do not include the compiled `.js`, `.js.map`, or `.d.ts` in your pull request as it makes it harder to merge your changes. 

But `master` contains these files for your main module. This confused me, so I removed the build artifacts to make the desired organization clearer.


domoritz:
Thank you for the PR. We want the compiled files in master but we don't want them in PRs to avoid merge conflicts. Does this make sense?

Issue #189, opened 2018-04-04
type Definition is missing property $schema


EvaLok:
i believe it just needs to be added here:
https://github.com/YousefED/typescript-json-schema/blob/master/typescript-json-schema.ts#L58

property `$schema` currently exists when generating a schema, but is not present in the type definition for `Definition`:

```json
{
  "type": "object",
  "properties": {
    "stringThing": {
      "type": "string"
    },
    "numberThing": {
      "type": "number"
    },
    "booleanThing": {
      "type": "boolean"
    }
  },
  "required": [
    "booleanThing",
    "numberThing",
    "stringThing"
  ],
  "$schema": "http://json-schema.org/draft-06/schema#"
}
```

EvaLok:
i think i found another thing, `properties` needs and index signature, so what i propose is:

```typescript
export type Definition = {
    $ref?: string,
    $schema ?: string; // <-- this
    description?: string,
    allOf?: Definition[],
    oneOf?: Definition[],
    anyOf?: Definition[],
    title?: string,
    type?: string | string[],
    definitions?: {[key: string]: any},
    format?: string,
    items?: Definition | Definition[],
    minItems?: number,
    additionalItems?: {
        anyOf: Definition[]
    },
    enum?: PrimitiveType[] | Definition[],
    default?: PrimitiveType | Object,
    additionalProperties?: Definition | boolean,
    required?: string[],
    propertyOrder?: string[],
    properties?: {[index: string]: string}, // <-- this
    defaultProperties?: string[],

    typeof?: "function"
};

```

domoritz:
Thank you for the issue. Can you send a PR?

EvaLok:
sure - should be able to submit one tomorrow

picheli20:
It can be closed then, no? https://github.com/YousefED/typescript-json-schema/commit/4cb6e5e900f1736f47c0f18c456161a532bc8dc4

PR #190, opened 2018-04-05
add `$schema` property to `Definition`, also add index sig to `proper…

EvaLok:
…ties`

fix: https://github.com/YousefED/typescript-json-schema/issues/189

Please:
- [ ] Do not include the compiled `.js`, `.js.map`, or `.d.ts` in your pull request as it makes it harder to merge your changes. 


EvaLok:
i find the first point in the list a bit confusing.. by putting those files in `.gitignore` and then having a `prepublishOnly` script call your build command, you could eliminate the need for having these files in the repository at all

PR #191, opened 2018-04-06
fix: not working example in README.txt

many20:


domoritz:
Hmm, Isn't the base path exactly there for this reason? Why is this change necessary?

domoritz:
Maybe there is a bug but the readme is correct.

Issue #154, opened 2017-10-18
Ignoring functions and constructor


KurtGokhan:
Currently, there seems no way to ignore functions. My class is:

``` typescript
export class Configuration {
  myProperty: MyPropertyType;
  toAnotherClass(): AnotherClass {  ...  }
}
```
I want `toAnotherClass` to be ignored. If there is a constructor, I also want to ignore it.

Also, can someone explain what is the use case for having functions in JSON schema? Can you define functions in JSON? For example, what is the benefit of having `toAnotherClass` in JSON schema as type `object` in the above example?

domoritz:
What does https://github.com/YousefED/typescript-json-schema/blob/master/typescript-json-schema.ts#L1050 do? I'd be happy to add a flag to ignore functions. There are very few use cases where it matters, imho.

KurtGokhan:
@domoritz I am also conflicted as I couldn't find any issues about ignoring functions. I thought, maybe there is a use case of having functions in JSON and I can't see it. But as I see it, I am going to use the generated schema on some `.json` file and it will not let me assign functions to a property anyway.

Using `typeof` makes it somehow better. The function is still in the schema but at least it is not object type which can confuse the the user of the schema. So instead of this:

```
                "toAnotherClass": {
                    "type": "object"
                },
```

You get this:

```
                "toAnotherClass": {
                    "typeof": "function"
                },
```

domoritz:
Another option is to define interfaces without functions and make your classes implement that interface. That's how we do things in (Vega-Lite)[https://github.com/vega/vega-lite]. We also use https://github.com/vega/ts-json-schema-generator but I just tested it and it crashes on function because it doesn't recognize the (ts) node type.

KurtGokhan:
Having an interface without functions sounds like a reasonable solution. I will go with that for now. Thank you.

domoritz:
Ignoring functions shouldn't be that hard either. If you create a test as a
pr in ts-json-schema-generator, I'll implement the feature.

KurtGokhan:
I have created the PR. #155 

Please tell me if the tests aren't enough.

domoritz:
Hey, sorry, I meant in https://github.com/vega/ts-json-schema-generator. Unfortunately, I don't have the time to add it to this repository unless you send a pull request with the implementation.

PR #194, opened 2018-04-11
Don't include methods

schani:
They're currently included as `object` types, which doesn't seem useful.


domoritz:
Fixes #110, fixes #154

Can you rebase this PR onto master? I will merge it then.

schani:
Done.

PR #195, opened 2018-04-11
Don't treat string enum members as numbers. Fixes #186

### LINKED ISSUE #186: String enum members starting with digits are translated into numbers
Input:

```
export enum Egg {
    Foo = "10 km",
    Bar = "2 km"
}
```

Output:

```
{
  "$schema": "http://json-schema.org/draft-06/schema#",
  "definitions": {
    "Egg": {
      "title": "Egg",
      "enum": [
        10,
        2
      ],
      "type": "number"
    }
  }
}
```

Weirdly, if only `Foo` is present, the `enum` is generated correctly.

schani:


domoritz:
Thank you!

PR #202, opened 2018-04-27
Cleanups

schani:


domoritz:
Thank you!

I noticed that the spacing is a bit inconsistent right now. Should we add prettier @schani?

schani:
I use prettier, so yes, I would appreciate that.

schani:
Please don't do this while you have PRs outstanding, as it will introduce painful merge conflicts.

PR #218, opened 2018-05-29
Support plugins

bergundy:
As discussed in #121 this is a proposal for adding plugin support for TJS.
I'll provide a test case and update the docs if we decide this is the right direction.
An example plugin implementing method schema generation can be seen here (code stolen from #121).

I removed the condition that returns null if property is a method, I'm guessing there are implications and there's a better way to do it.

Note: it's better to view the diff with with whitespace ignored.

### DIFF
--- a/typescript-json-schema-cli.ts
+++ b/typescript-json-schema-cli.ts
@@ -38,6 +38,8 @@ export function run() {
             .describe("uniqueNames", "Use unique names for type symbols.")
         .array("include").default("*", defaultArgs.include)
             .describe("include", "Further limit tsconfig to include only matching files.")
+        .array("plugins").default("plugins", defaultArgs.plugins)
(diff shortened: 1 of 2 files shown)


domoritz:
This looks good to me if you can get the tests to pass.

bergundy:
@domoritz tests pass, I added another flag `--filterMethods` to make them pass

domoritz:
Looks great! Let's see what @otgerrogla says.

bergundy:
I've decided to abort this PR since I found a better way to support methods in TJS.
Instead of copying the code from #121 which produces a schema which is not JSON schema compatible I'm describing methods in this manner:

```
interface Foo {
  bar: {
    params: {
       a: number;
    };
    returns: string;
  };
}
```

Later I might use a tool that converts an interface method signature into the structure above.

Thanks for the time put in this.

PR #220, opened 2018-06-20
buildGenerator performance improvement (avoid unnecessary work)

curledUpSheep:
- don't get type checker if we will not use it
- don't get diagnostics if we will ignore them anyway

This performance improvement is relevant for my programmatic use case.

Note: I got 1 failed test both before and after my change. Details:

  1) schema
       other
         unique-names should create correct schema:
     Error: ENOENT: no such file or directory, open 'C:\Users\********\git\typesc
ript-json-schema\test\programs\unique-names\schema.MyObject.2219977.json'
      at Object.fs.openSync (fs.js:646:18)
      at Object.fs.readFileSync (fs.js:551:33)
      at Context.<anonymous> (test\schema.test.ts:55:24)

Please:

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -1029,11 +1029,14 @@ export function buildGenerator(program: ts.Program, args: PartialArgs = {}): Jso
         }
     }
 
-    const typeChecker = program.getTypeChecker();
+    let diagnostics: Array<ts.Diagnostic> = [];
 
-    var diagnostics = ts.getPreEmitDiagnostics(program);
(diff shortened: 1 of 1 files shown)


domoritz:
Lgtm

curledUpSheep:
Seems to me like Travis was under high load causing the first test to take slightly longer than the 5000ms timeout. Will try to trigger a rebuild by closing and reopening this PR.

curledUpSheep:
Ok, that has done the trick :) tests ran a lot faster this time

PR #222, opened 2018-07-01
Really include only requested files

bergundy:
This is better than what we had before, now `tjs` will prefer included files over a project dependency.
It's still not a perfect solution though.


domoritz:
Can you explain the issue that this is addressing? Thanks!

domoritz:
Thank you for the patch!

PR #223, opened 2018-07-05
Support TypeScript 2.9.x

rpetrich:
Update the dependency and fix resulting issues related to type formatting


domoritz:
Can you resolve the issues and fix the build? It'd be great to merge this.

rpetrich:
@domoritz this should be ready to go.

I had to introduce a symbol hashing function to make the `unique-names` test deterministic as Travis was getting different symbol IDs than my local installs.

Also, we may want to update the tsnode dependencies now that the bundled TypeScript is 2.9.x

rpetrich:
@domoritz ts-node is now updated to 7.0.0

domoritz:
Thank you!

PR #224, opened 2018-07-05
Add rejectDateType option

rpetrich:
Add a `--rejectDateType` option that causes typescript-json-schema to reject all use of `Date` in schemas


rpetrich:
Validating a value against a schema produced from a interface `T` ensures that the value conforms to `T` and that the value can be safely cast to `T` in TypeScript. This holds true for all of the types that typescript-json-schema supports except for `Date`. Fields of the type `Date` allow strings to validate, which means the value can't be safely cast to `T` as the value of those fields will be strings and not `Date`s. Rejecting `Date` types upfront prevents the generation of an overly permissive schema.


The context for this is I use typescript-json-schema in a module system that automatically generates validator functions (via ajv) for TypeScript interfaces. This makes it easy to check the runtime types of values pulled from a REST API or untrusted peers and fail fast if they don't match:

module: `point.ts`
```typescript
export interface Point {
    x: number;
    y: number;
}
```

generated pseudo-module: `point.ts!validators`
```typescript
import { Point as PointType } from "./point.ts";
export function Point(value: any): value is PointType {
    // generated code goes here
}
```

consuming module:
```typescript
import { Point as isPoint } from "./point.ts!validators";
const value = await fetchFromAPIAsJSON("...");
if (isPoint(value)) {
    console.log(`x: ${value.x}, y: ${value.y}`);
} else {
    throw new Error("Expected a Point, got something else!");
}
```

domoritz:
Thank you!

Issue #183, opened 2018-02-23
feature request: support $id keyword


marcelh-gh:
**feature request:**
It would be nice if a value for the $id keyword could be assigned via the options.

**motivation:**
When using multiple schemas together (for example one main one with definitions within others), validators like AJV are using the $id keyword for referencing between the schemas.

marcelh-gh:
I have just seen that this is probably possible with the `validationKeywords `option. maybe an example for this would be nice.

domoritz:
Note that draft 06 uses `id` instead.

domoritz:
Can you send a PR?

handrews:
@domoritz draft-06 uses `$id` and is the draft that renamed draft-04's `id` to `$id`.  I'm quite certain of this as I'm the person who did it :-)

domoritz:
Ups, I mixed up the two. Thanks for the clarification  @handrews!

beeing:
Before putting in a PR, is this what we are looking at?

- Add arg like `--id schema`
- Generate the schema with `"$id": "schema"` 
- Each $refs will contain the prefix of schema as in `schema#/definitions/typename`

PR #229, opened 2018-07-17
added argument for $id in schema object

lumonad:
I'm going to use [ajv](https://github.com/epoberezkin/ajv) as validator, and for proper work of references it requires schema `$id`

I think it's what somebody wants here https://github.com/YousefED/typescript-json-schema/issues/183

### DIFF
--- a/test/programs/argument-id/schema.MyObject.json
+++ b/test/programs/argument-id/schema.MyObject.json
@@ -1,0 +1,33 @@
+{
+    "$id": "someSchemaId",
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "definitions": {
+        "ReferenceType": {
+            "properties": {
+                "reference": {
+                    "enum": [
+                        true
+                    ],
+                    "type": "boolean"
+                }
+            },
+            "required": [
+                "reference"
+            ],
+            "type": "object"
+        }
+    },
+    "properties": {
+        "referenceType": {
+            "$ref": "someSchemaId#/definitions/ReferenceType"
+        },
+        "someProp": {
+            "type": "string"
+        }
+    },
+    "required": [
+        "referenceType",
+        "someProp"
+    ],
+    "type": "object"
+}
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -29,6 +29,7 @@ export function getDefaultArgs(): Args {
         include: [],
         excludePrivate: false,
         uniqueNames: false,
+        id: ""
     };
 }
 
@@ -53,6 +54,7 @@ export type Args = {
     include: string[];
     excludePrivate: boolean;
     uniqueNames: boolean;
+    id: string;
 };
 
 export type PartialArgs = Partial<Args>;
@@ -62,6 +64,7 @@ export type PrimitiveType = number | boolean | string | null;
 export type Definition = {
     $ref?: string,
     $schema?: string,
+    $id?: string,
     description?: string,
     allOf?: Definition[],
     oneOf?: Definition[],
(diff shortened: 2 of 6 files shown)


typescript-json-schema.ts
```diff
@@ -859,7 +862,7 @@ export class JsonSchemaGenerator {
             // We don't return the full definition, but we put it into
             // reffedDefinitions below.
             returnedDefinition = {
-                $ref:  "#/definitions/" + fullTypeName
+                $ref:  `${this.args.id}#/definitions/` + fullTypeName
```
domoritz: Can you show me an example of this used somewhere?

lumonad: if I understand your question correct, it's how ajv expects to work with refs https://github.com/epoberezkin/ajv#ref

domoritz: Is this in the JSON Schema standard?

domoritz:
Looks good to me. Please add a test and provide a link to the json schema standard and I will merge. Thank you for the contribution @lumonad!

typescript-json-schema.ts
lumonad: I think so) http://json-schema.org/latest/json-schema-core.html#rfc.section.8.2

domoritz: But this is to access things from another schema, right? Why would I want to add an id to refer to stuff within the same file?

lumonad:
I almost fixed tests, but it keep failing on `unique-names`. And I'm not sure that's because of my changes. Just verified this test case failed on master as well, do you know how to solve it?

domoritz:
Master builds fine: https://travis-ci.org/YousefED/typescript-json-schema/builds/404033154

test/programs/argument-id/schema.MyObject.json
```diff
@@ -0,0 +1,33 @@
...
+            },
+            "required": [
+                "reference"
+            ],
+            "type": "object"
+        }
+    },
+    "properties": {
+        "referenceType": {
+            "$ref": "someSchemaId#/definitions/ReferenceType"
```
domoritz: Doesn't someSchemaId have to be a URL? I'd like this to be a somewhat plausible example.

typescript-json-schema.ts
lumonad: It's a good question for owner of that library. From what I saw they relay on this id as key in object of multiple schemas. And without such `$id` property they just set schema to empty string key, and it doesn't work properly.

I can describe whole use case where it broken. When I specify file path pattern through --include, it puts all my actual interfaces as definitions, so I don't have some root schema and only a bunch of definitions.

for example:
``` typescript
// file interfaces/Interface1.root.ts
import SubInterface1 from '../shared/SubInterface1';
interface Interface1 {
  prop1: SubInterface1;
}

// file interfaces/Interface2.root.ts
import SubInterface1 from '../shared/SubInterface1';
interface Interface2 {
  prop1: SubInterface1
}

// file shared/SubInterface1.ts
interface Interface1 {
  prop1: SubInterface1
}
```

after that with command like `typescript-json-schema tsconfig.json "*" --include=interfaces/*.ts`

we will get a schema of definitions without any root object.

after that I set my schema for `ajv` like
``` typescript
import schema from 'schema.ts';
ajv.addSchema(schema);
...
```

and when I want to validate some specific data I put concrete part of this file:
```typescript
...
ajv.validate(schema.definitions.Interface1, objectToValidate1)
ajv.validate(schema.definitions.Interface2, objectToValidate2)
```

domoritz:
Thank you!

PR #230, opened 2018-07-23
Resolve type aliases

kayahr:
In some situations getSymbolAtLocation() returns an alias which does
not provide JSDoc comments. In this case the type must be resolved
with getAliasedSymbol first to be able to read comments for description
and annotations.

I also had to rename two schema files in the unique-names unit tests
because the symbol IDs are now different. I guess that's because now the
real symbols are used instead of the alias symbols. I hope this isn't
a breaking change, I'm not even sure what the use case for this unique
name feature is anyway.

Please:
- [X] Do not include the compiled `.js`, `.js.map`, or `.d.ts` in your pull request as it makes it harder to merge your changes. 


domoritz:
I think changing the id of objects is not a breaking change since noone should depend on it.

domoritz:
Also, we are pre 1.0 so everything is fair game. I released 0.28

PR #234, opened 2018-07-30
TypeScript 3.0 support

rpetrich:
Update to TypeScript 3.0 and support the variable-length tuples and rest parameter tuples


domoritz:
Thank you!

PR #235, opened 2018-07-30
Support TypeScript's ReadOnly and Mutable types

### LINKED ISSUE #181: Readonly<> declarations are not properly handled
test.ts:
```
type Test = Readonly<{ test: string }>
```

output:
```
typescript-json-schema test.ts --required Test
{
    "properties": {
    },
    "type": "object"
}
```

I guess the readonly should just be stripped, so the expected output would be something like:
```
{
    "properties": {
        "test": {
            "type": "string"
        }
    },
    "required": [
        "test"
    ],
    "type": "object"
}
```

rpetrich:
Treat them as passthrough since they only add modifiers to the types of fields.

Fixes #181


typescript-json-schema.ts
```diff
@@ -821,6 +821,11 @@ export class JsonSchemaGenerator {
     private getTypeDefinition(typ: ts.Type, asRef = this.args.ref, unionModifier: string = "anyOf", prop?: ts.Symbol, reffedType?: ts.Symbol, pairedSymbol?: ts.Symbol): Definition {
         const definition: Definition = {}; // real definition
 
+        while (typ.aliasSymbol && (typ.aliasSymbol.escapedName === "Readonly" || typ.aliasSymbol.escapedName === "Mutable") && typ.aliasTypeArguments && typ.aliasTypeArguments[0]) {
```
domoritz: Can you add a comment here why we are ignoring these types?

PR #240, opened 2018-08-22
Throw errors when problems happen so we get non-zero exit codes

### LINKED ISSUE #239: When tool fails with "Unable to write output file" exit with non-zero exit code
Right now, even though it failed (in my particular case with `Unable to write output file: EACCES: permission denied, open 'scenarioConfig.json'`) it exits with code 0, which makes it hard to work with in CI/programmatic environments. Can it please be changed to have a non-zero exit code?

palfrey:
Please:

Fixes #239

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -1195,14 +1195,14 @@ export function exec(filePattern: string, fullTypeName: string, args = getDefaul
 
     const definition = generateSchema(program, fullTypeName, args, onlyIncludeFiles);
     if (definition === null) {
-        return;
+        throw new Error("No output definition. Probably caused by errors prior to this?");
     }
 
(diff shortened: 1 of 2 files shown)


typescript-json-schema.ts
```diff
@@ -1195,14 +1195,14 @@ export function exec(filePattern: string, fullTypeName: string, args = getDefaul
 
     const definition = generateSchema(program, fullTypeName, args, onlyIncludeFiles);
     if (definition === null) {
-        return;
+        throw new Error("No output definition. Probably caused by errors prior to this?");
```
domoritz: Can you give an example that triggers this line? I want to make sure that we want to raise an error here.

palfrey: Added a test case that does so

PR #247, opened 2018-10-05
Consider methods when typeofKeyword is true

marionebl:
This changes the current behaviour of always ignoring method declarations to include them via `{ "typeof": "function" }` when `args.typeofKeyword` is true.

I wonder if handling function properties and methods the same would be the best way forward. Let me know.


domoritz:
Closing since it's outdated by now.

Issue #245, opened 2018-09-17
Schema for object with number index


CaselIT:
The typescript type
```ts
interface MyType {
  [index: number]: number
}
```
is parsed into the schema
```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "items": {
        "type": "number"
    },
    "type": "array"
}
```
while it should be parsed into
```
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "type": "object",
    "patternProperties": {
        "^[0-9]+$": {
            "type": "number"
        }
    },
    "additionalProperties": false
}
```

domoritz:
Yeah, that is weird. Can you send a PR to fix it?

CaselIT:
I can try, but I do not know when I will have the time

domoritz:
Thanks!

CaselIT:
Do you think the current behaviour should be kept under a switch?

domoritz:
It looks incorrect to me. If we don't have a test for it, probably no.

CaselIT:
I've found the time to look into fixing this. I've added a pull request that fixes this

PR #248, opened 2018-10-29
Correct schema for object with number index

CaselIT:
This pull request fixes the issue #245 

It's the first time that I use the typescript api in a program so probably there is a better solution.

Please:


CaselIT:
Thanks for the quick update @domoritz

domoritz:
Thanks for the PR!

PR #254, opened 2018-12-24
resolved issue where schema gen would crash when using abstract classes

garyevari:
Please:
- [ ] Do not include the compiled `.js`, `.js.map`, or `.d.ts` in your pull request as it makes it harder to merge your changes. 

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -723,7 +723,7 @@ export class JsonSchemaGenerator {
 
         const modifierFlags = ts.getCombinedModifierFlags(node);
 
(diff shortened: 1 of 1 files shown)


domoritz:
Can you add a small test case so we make sure not to have regressions in the future?

garyevari:
> Can you add a small test case so we make sure not to have regressions in the future?

I will have to do another PR first to fix typescript errors - master cannot be built.

garyevari:
@domoritz tests have been added for abstract class use case

there is still 2 tests failing, but they are also failing in master
1) type-aliases should create correct schema
2) type-aliases-recursive-object-topref should create correct schema

Cheers

PR #261, opened 2019-02-09
Ignore declarationMap option in TypeScript config.

plampila:
Using declarationMap option without declaration option (which is already ignored) results in error:
Option 'declarationMap' cannot be specified without specifying option 'declaration' or option 'composite'.


domoritz:
Thank you! Can you provide the command that causes the error?

plampila:
Just: typescript-json-schema tsconfig.json AnyType

When tsconfig.json includes both declaration and declarationMap options (and I guess no composite option).

domoritz:
Thanks

PR #264, opened 2019-02-14
Fix an error where two different partial types get confused

ForbesLindesay:
Both these two types get the alias `Partial`, even though they have different strucutres.

Please:


domoritz:
Thanks

PR #268, opened 2019-02-28
Add support for TypeScript project references

kasperisager:
This change ensures that project references are passed along when creating a TypeScript program.


domoritz:
Thanks!

PR #270, opened 2019-03-06
Advanced types - Exclude<T, S> should generate separate types in schema, if the arguments are different.

leqwasd:
Advanced types - Exclude<T, S> should generate separate types in schema, if the arguments are different.

I just added here a test case.
For interface:

    export interface IExcludeTest {
        AB: IExcludeTestAB;
        CD: IExcludeTestCD;
    }

    export interface IExcludeTestAB {
        A: string;
        B: string;
        importantKey: Exclude<keyof IExcludeTestAB, "importantKey">;
    }
    export interface IExcludeTestCD {
        C: string;
        D: string;
        importantKey: Exclude<keyof IExcludeTestCD, "importantKey">;
    }

It will generate one type

    "Exclude": {
        "description": "Exclude from T those types that are assignable to U",
        "enum": [
            "A",
            "B"
        ],
        "type": "string"
    }

And reference it in both situations

    "importantKey": {
        "$ref": "#/definitions/Exclude"
    }

### DIFF
--- a/test/programs/exclude-unique-type/main.ts
+++ b/test/programs/exclude-unique-type/main.ts
@@ -1,0 +1,15 @@
+export interface IExcludeTest {
+    AB: IExcludeTestAB;
+    CD: IExcludeTestCD;
+}
+
+export interface IExcludeTestAB {
+    A: string;
+    B: string;
+    importantKey: Exclude<keyof IExcludeTestAB, "importantKey">;
+}
+export interface IExcludeTestCD {
+    C: string;
+    D: string;
+    importantKey: Exclude<keyof IExcludeTestCD, "importantKey">;
+}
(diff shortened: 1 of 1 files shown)


domoritz:
Is the issue the exclude or can we see the same issue with generics?

leqwasd:
I checked for these:

> Predefined conditional types
> TypeScript 2.8 adds several predefined conditional types to lib.d.ts:
> Exclude<T, U> – Exclude from T those types that are assignable to U.
> Extract<T, U> – Extract from T those types that are assignable to U.
> NonNullable<T> – Exclude null and undefined from T.
> ReturnType<T> – Obtain the return type of a function type.
> InstanceType<T> – Obtain the instance type of a constructor function type.

And found this problem with Exclude and Extract

Checked for NonNullable and ReturnType aswell, but they worked as expected.

domoritz:
What about a custom generic?

PR #271, opened 2019-03-11
feat: 🎸 allow namespaced versions of "Date"

johngeorgewright:
I'm trying to use this package to build schemas that include [gapi.client.people](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/gapi.client.people) types.

[They use their own `Date` type](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6a5529e838d9386fb6567296e84d945e25805281/types/gapi.client.people/index.d.ts#L182-L195):

```typescript
        interface Date {
            /**
             * Day of month. Must be from 1 to 31 and valid for the year and month, or 0
             * if specifying a year/month where the day is not significant.
             */
            day?: number;
            /** Month of year. Must be from 1 to 12. */
            month?: number;
            /**
             * Year of date. Must be from 1 to 9999, or 0 if specifying a date without
             * a year.
             */
            year?: number;
        }
```

This does not build with typescript-json-schema, as it considers any interface with a name of `Date` to be a native Date.

This pull request will look for a "full qualified name" of `Date` instead.

### DIFF
--- a/test/programs/custom-dates/main.ts
+++ b/test/programs/custom-dates/main.ts
@@ -1,0 +1,11 @@
+namespace foo {
+  export interface Date {
+    day?: number;
+    month?: number;
+    year?: number;
+  }
+
+  export interface Bar {
+    date: Date;
+  }
+}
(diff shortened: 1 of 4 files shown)


johngeorgewright:
Hi @YousefED. Any thoughts on this?

domoritz:
This is great. Thanks!

PR #272, opened 2019-04-04
Update to TypeScript 3.4

marionebl:
Update to TypeScript 3.4. This includes the following the changes. 

* Update specified TypeScript dependency to `3.4`
* Update yarn.lock to format > 1.9, See [yarnpkg/yarn#6440](https://github.com/yarnpkg/yarn/issues/6440)
* Simplify a condition in `getTypeDefinition` that was not entered after upgrading to `3.4`, which causes some tests regarding aliases to fail. I cross checked with `3.0` and the proposed changes work with those older versions, too. 

<details>
Please:
- [] ~~Provide a test case & update the documentation in the `Readme.md`~~ Not applicable
</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
   "dependencies": {
     "glob": "~7.1.2",
     "json-stable-stringify": "^1.0.1",
-    "typescript": "^3.0.1",
+    "typescript": "^3.4.1",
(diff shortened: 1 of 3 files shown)


domoritz:
Thank you. Looks great.

PR #278, opened 2019-05-15
Add support for unknown

ForbesLindesay:
The `unknown` keyword is like a safe version of `any`. It does not verify anything about the value, but acts unlike `any`, it is a type error to assume anything about the value of an `unknown`.

Please:


domoritz:
Thank you

Issue #284, opened 2019-05-27
Failing tests


HoldYourWaffle:
I was planning on investigating a possible fix for #283, but sadly the tests fail without me changing anything.

```
82 passing (2m)
  5 failing

  1) schema
       type aliases
         type-alias-single-annotated should create correct schema:

      The schema is not as expected
      + expected - actual

       {
         "$schema": "http://json-schema.org/draft-07/schema#"
      +  "description": "This is a description"
      +  "maxLength": 24
      +  "minLength": 10
      +  "pattern": "^mystring-[a-zA-Z0-9]+$"
         "type": "string"
       }

      at Context.<anonymous> (test\schema.test.ts:29:16)

  2) schema
       unions and intersections
         type-intersection-recursive should create correct schema:

      The schema is not as expected
      + expected - actual

       {
         "$schema": "http://json-schema.org/draft-07/schema#"
      -  "definitions": {}
      +  "definitions": {
      +    "ChildFoo": {
      +      "type": "object"
      +    }
      +    "Foo": {
      +      "properties": {
      +        "childFoos": {
      +          "allOf": [
      +            {
      +              "$ref": "#/definitions/Foo"
      +            }
      +            {
      +              "$ref": "#/definitions/ChildFoo"
      +            }
      +          ]
      +        }
      +      }
      +      "required": [
      +        "childFoos"
      +      ]
      +      "type": "object"
      +    }
      +  }
       }

      at Context.<anonymous> (test\schema.test.ts:29:16)

  3) schema
       other
         generate-all-types should create correct schema:

      The schema is not as expected
      + expected - actual

       {
         "$schema": "http://json-schema.org/draft-07/schema#"
      -  "definitions": {}
      +  "definitions": {
      +    "MyEnum": {
      +      "enum": [
      +        0
      +      ]
      +      "type": "number"
      +    }
      +    "MyInterface": {
      +      "type": "object"
      +    }
      +    "MyObject": {
      +      "type": "object"
      +    }
      +  }
       }

      at Context.<anonymous> (test\schema.test.ts:29:16)

  4) schema
       other
         unique-names should create correct schema:
     Error: ENOENT: no such file or directory, open 'test/programs/unique-names/schema.MyObject.6aa2b6a1.json'
      at Object.openSync (fs.js:436:3)
      at Object.readFileSync (fs.js:341:35)
      at Context.<anonymous> (test\schema.test.ts:53:24)

  5) schema
       other
[...]

PR #285, opened 2019-05-27
Add v10 to travis

HoldYourWaffle:
Investigating #284.
I just want to check if Travis still passes if it runs on node 10 too, since that's what I'm using.


domoritz:
Looks like it's good. I don't think we need to run tests on multiple node versions. In my experience, I've never had an issue that one is one version of node and not the other.

PR #286, opened 2019-05-28
Asterisk to include all files

tuliodiaz:
Please update the docs. I lost three hours because the * need to be '*'.
If you run the command with no quotes its try to scan the current folder where you execute the command.


domoritz:
I'm sorry to hear about your wasted time. The specific behavior for this case depends on your shell but you are absolutely right that in most shells the * should be escaped. Thanks for the PR!

PR #294, opened 2019-05-31
Ignore compiled JS files

HoldYourWaffle:
- [x] Do not include the compiled `.js`, `.js.map`, or `.d.ts` in your pull request as it makes it harder to merge your changes (this is the main reason for the PR)
- [x] Provide a test case & update the documentation in the `Readme.md` (not applicable)

I'm pretty sure there's no reason to include the compiled JS files in the repository. Users can build the files themselves if they need to (although they are of course also available from NPM).
There are a lot of cons to including compiled files, including but not limited to:

1. Big source of conflicts (it's even mentioned in the PR template)
2. Code duplication (it's completely redundant)
3. Diff clutter, we're (almost) never interested in differences between compiled code
4. Filesystem clutter. Having 3-4 files with almost the same name hurts readability a lot. I solved this by adding an `outDir` to `tsconfig.json` set to `dist/` as per common convention.

The compiled files are of course still published to NPM.


tsconfig.json
```diff
@@ -2,7 +2,8 @@
     "compilerOptions": {
         "target": "es5",
         "module": "commonjs",
-        "moduleResolution": "node",
+		"moduleResolution": "node",
```
domoritz: Fix indentation

package.json
```diff
@@ -65,7 +65,7 @@
     "tslint": "^5.16.0"
   },
   "scripts": {
-    "test": "npm run build && mocha -t 5000 --require source-map-support/register test",
+    "test": "npm run build && mocha -t 5000 --require source-map-support/register dist/test",
     "debug": "ts-node --inspect=19248 --debug-brk typescript-json-schema-cli.ts",
     "run": "ts-node typescript-json-schema-cli.ts",
     "build": "tsc -p .",
```
domoritz: needs a `prepublishOnly `

tsconfig.json
domoritz: We use spaces

HoldYourWaffle: I blame VS Code. Fixed now.

domoritz: Thanks

package.json
HoldYourWaffle: What do you mean? Do you want to run the build on `prepublishOnly`?

domoritz: We want to run the build. See https://github.com/vega/ts-json-schema-generator/blob/master/package.json#L61

HoldYourWaffle: Added

domoritz:
Thank you

PR #296, opened 2019-06-03
Fix the executable issue

YuJianrong:
The compiled file `typescript-json-schema-cli.js` moved to `/dist/` now.

Please:


domoritz:
Thanks

PR #303, opened 2019-07-18
add support for wildcard validationKeyword annotations. support ui:*

mengwong:
Support annotations of the form `@TJS-ui:foo bar`. This is useful for downstream extraction to systems like react-jsonschema-form.

add examples for `ui:autofocus true`

add documentation to README.md

Please:

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ export interface Shape {
      *
      * @minimum 0
      * @TJS-type integer
+     * @TJS-ui:autofocus true
      */
     size: number;
 }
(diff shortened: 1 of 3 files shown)


domoritz:
This seems very specific for one use case. Can we just allow any annotation instead?

mengwong:
Do you mean, any `@TJS-XXX YYY` should transform into `XXX: YYY` ? I could resubmit a PR that does that.

domoritz:
That sounds good. I think we used to have that at some point, actually.

domoritz:
Please submit a new pull request.

Issue #310, opened 2019-08-13
Missing properties in `Definition`


JemarJones:
I may be missing something obvious here, but the `Definition` typescript type seems to be missing keys which are a part of the json schema standard. For example, [dependencies](https://json-schema.org/understanding-json-schema/reference/object.html#dependencies) or the [not](https://json-schema.org/understanding-json-schema/reference/combining.html#not) keyword.

domoritz:
They may be part of the latest draft and not yet implemented. It would be good to switch to import this type from `@types/json-schema`. Can you send a PR?

PR #311, opened 2019-08-15
Use @types/json-schema definition for more standard json schema Defin…

JemarJones:
…ition

Please:
^ Not sure this is relevant, all the existing tests are passing.

----
Based on discussion in https://github.com/YousefED/typescript-json-schema/issues/310 with @domoritz, i've replaced the existing json schema `Definition` with one based on the `@types/json-schema` `JsonSchema7` definition. Rather than use this type verbatim though, i've extended it to do the following:
- Keep the 3 existing fields which appear to be non-standard `propertyOrder`, `defaultProperties`, and `typeof`.
- Maintain the existing definition of `type`. The issue here is that in addition to the standard types (`'string' | 'number' | 'integer' | 'boolean' | 'object' | 'array' | 'null'`), the value of `undefined` is being used as well. The json-schema doesn't definition doesn't allow for this (as i believe it's mostly equivalent to simply omitting the property from the `required` properties list?), but i've extended the definition in such a way to allow this as is. Might be worth considering properly fixing this issue in the future.
- I also had to redefine all of the nested fields which referenced the `JSONSchema7` definition itself in some way, to correctly use this extended definition throughout.

Ran the test suite and had no issues. I'm not intimately familiar with this project, so apologise if anything has been missed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "schema"
   ],
   "dependencies": {
+    "@types/json-schema": "^7.0.3",
     "glob": "~7.1.4",
     "json-stable-stringify": "^1.0.1",
     "typescript": "^3.5.1",
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -70,12 +70,12 @@ type RedifinedFields = "type" | "items" | "additionalItems" | "contains" | "prop
 export type DefinitionOrBoolean = Definition | boolean;
 export interface Definition extends Omit<JSONSchema7, RedifinedFields> {
     // The type field here is incompatible with the standard definition
-    type?: string | string[],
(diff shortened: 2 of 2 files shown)


domoritz:
Thank you!

I am obviously not too happy about our use of "undefined". It would be great to fix this since we should just remove the property from `required` instead. Could you look into this?

Another thing we should do now is to make `validationKeywords` safer. If we make it of type `{[K in Definition]: true}`, then we will get a compile-time error when a property is missing. Agian, I would love a follow up PR towards this goal.

JemarJones:
No problem :). I will try to look into those when i have some time, may take me a little bit to get more familiar with the relevant code!

PR #312, opened 2019-08-16
Remove "undefined" usage with definition type

JemarJones:
Please:

----

As discussed in https://github.com/YousefED/typescript-json-schema/pull/311#issuecomment-521566082, removed usages of `"undefined"` definition type so that the the type property in `Definition` could be corrected.  The existing test `strict-null-checks should create correct schema` already describes the correct behaviour of removing prop from the required array, and this test continues to pass with these updates.

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -3,7 +3,7 @@ import * as stringify from "json-stable-stringify";
 import * as path from "path";
 import { createHash } from "crypto";
 import * as ts from "typescript";
-import { JSONSchema7 } from "json-schema";
+import { JSONSchema7, JSONSchema7TypeName } from "json-schema";
 export { Program, CompilerOptions, Symbol } from "typescript";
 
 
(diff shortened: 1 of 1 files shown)


typescript-json-schema.ts
```diff
@@ -66,12 +66,9 @@ export type PartialArgs = Partial<Args>;
...
 
-type RedifinedFields = "type" | "items" | "additionalItems" | "contains" | "properties" | "patternProperties" | "additionalProperties" | "dependencies" | "propertyNames" | "if" | "then" | "else" | "allOf" | "anyOf" | "oneOf" | "not" | "definitions";
+type RedifinedFields = "items" | "additionalItems" | "contains" | "properties" | "patternProperties" | "additionalProperties" | "dependencies" | "propertyNames" | "if" | "then" | "else" | "allOf" | "anyOf" | "oneOf" | "not" | "definitions";
 export type DefinitionOrBoolean = Definition | boolean;
 export interface Definition extends Omit<JSONSchema7, RedifinedFields> {
-    // The type field here is incompatible with the standard definition
-    type?: string | string[];
-
     // Non-standard fields
     propertyOrder?: string[];
```
domoritz: We should probably get rid of these non-standard properties at some point as well. Not in this PR, though.

typescript-json-schema.ts
```diff
@@ -196,7 +193,7 @@ const simpleTypesAllowedProperties = {
     description: true
 };
 
-function addSimpleType(def: Definition, type: string) {
+function addSimpleType(def: Definition, type: JSONSchema7TypeName) {
```
domoritz: Beautiful

typescript-json-schema.ts
```diff
@@ -607,9 +604,10 @@ export class JsonSchemaGenerator {
                 pushEnumValue(value);
             } else {
                 const def = this.getTypeDefinition(valueType);
-                if (def.type === "undefined") {
+                if ((<any>def).mayBeUndefined) {
```
domoritz: Can't say I'm a fan of this. 

If I understand correctly, we only want to make sure that we update the `required` property correctly, right? SO I think it really only matters in https://github.com/YousefED/typescript-json-schema/blob/921dc104f5fc3231a3e32b34585f05a04e08480a/typescript-json-schema.ts#L816. 

As a proposal, can you assemble the required properties you go through them in `getClassDefinition` and then in `getUnionDefinition` just look at the `required` property? This would avoid using `any`.

domoritz:
Will you get to address my comments?

JemarJones:
Yeup sorry, I will! Just have to find a bit of time.

domoritz:
Closing due to inactivity. Please open a new pull request.

PR #313, opened 2019-08-21
Fix Error @TJS-custom annotation without value

ndunks:
Treat it as boolean with default true

Please:

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -380,7 +380,7 @@ export class JsonSchemaGenerator {
                     name = match[1];
                     text = match[2];
                 } else {
-                    // Treat empty text as boolean
+                    // Treat empty text as boolean true
                     name = (text as string).replace(/^[\s\-]+/, "");
                     text = "true";
                 }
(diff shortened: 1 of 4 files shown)


typescript-json-schema.ts
```diff
@@ -373,7 +373,18 @@ export class JsonSchemaGenerator {
...
             // if we have @TJS-... annotations, we have to parse them
-            const [name, text] = (doc.name === "TJS" ? new RegExp(REGEX_TJS_JSDOC).exec(doc.text!)!.slice(1,3) : [doc.name, doc.text]) as string[];
+            let [ name, text ] = [doc.name, doc.text as string];
+            if( doc.name === "TJS" ) {
+                let match: string[] | RegExpExecArray | null = new RegExp(REGEX_TJS_JSDOC).exec(doc.text!);
+                if( match ) {
+                    name = match[1];
+                    text = match[2];
+                } else {
+                    // Treat empty text as boolean
```
domoritz:
```suggestion
                    // Treat empty text as boolean true
```

PR #314, opened 2019-08-22
Argument noschemaprop

ndunks:
Please:

Add ability to not creating `$schema` field, because when combining with `openapi`, I got error 'Should NOT have additional property: `$schema`'

image


domoritz:
Do you use the cli directly? Otherwise it'd be easy to drop the property from your json outside this tool.

Issue #236, opened 2018-08-01
Validation keywords for primitives in arrays


grigasp:
Would be nice if we could specify validation keywords for primitive array items. Example:
```
export interface Test {
  /**
   * @multipleOf 10
   */
  numbersArray?: number[];
}
```
now converts to
```
"Test": {
  "additionalProperties": false,
  "properties": {
    "numbersArray": {
      "items": {
        "type": "number"
      },
      "multipleOf": 10,
      "type": "array"
    },
  },
  "type": "object"
}
```
but should instead be converted to
```
"Test": {
  "additionalProperties": false,
  "properties": {
    "numbersArray": {
      "items": {
        "type": "number",
        "multipleOf": 10
      },
      "type": "array"
    },
  },
  "type": "object"
}
```

domoritz:
That would not be correct. 

Can you make a new type that just aliases number and add the comment to that type?

grigasp:
It doesn't look like TJS picks up type annotations for properties:
```
/**
 * Some type description
 * @multipleOf 10
 * @type integer
 */
export type SomeNumber = number;

export interface Test {
  /** 
   * Some property description
   */
  numbersArray?: SomeNumber[];
}
```
gets converted to
```
"Test": {
  "additionalProperties": false,
  "properties": {
    "numbersArray": {
      "description": "Some property description",
      "items": {
        "type": "number"
      },
      "type": "array"
    }
  },
  "type": "object"
}
```

domoritz:
Can you try https://github.com/vega/ts-json-schema-generator and see whether it handles aliases better?

grigasp:
Yes, it created a separate definition for the type alias. It's result:
```
"SomeNumber": {
  "description": "Some type description",
  "multipleOf": 10,
  "type": "number"
},
"Test": {
  "additionalProperties": false,
  "properties": {
    "numbersArray": {
      "description": "Some property description",
      "items": {
        "$ref": "#/definitions/SomeNumber"
      },
      "type": "array"
    }
  },
  "type": "object"
}
```

domoritz:
Aliases are difficult to handle with the approach in this library. Ts-json-schema-generator may be a better fit.

dobesv:
The original issue was about specifying information about the elements of an array.  I am trying to do this as well.

I tried an approach like this:

```
/**
 * @items {"type":"integer"}
 */
 grades: number[];
```

but it doesn't seem to set `items` for some reason.

PR #327, opened 2019-12-04
Allow "items" to be overridden on arrays using jsdoc annotations

dobesv:
I wanted to able to override the `items` of an array type using jsdoc annotations.  This PR makes that possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,65 @@ will be translated to
 
 Note that we needed to use `@TJS-type` instead of just `@type` because of an [issue with the typescript compiler](https://github.com/Microsoft/TypeScript/issues/13498).
 
+You can also override the type of array items, either listing each field in its own annotation or one
+annotation with the full JSON of the spec (for special cases). This replaces the item types that would
+have been inferred from the TypeScript type of the array elements.  
+
+Example:
+
+```ts
+export interface ShapesData {
+    /**
+     * Specify individual fields in items.
+     *
+     * @items.type integer
+     * @items.minimum 0
+     */
+    sizes: number[];
+
+    /**
+     * Or specify a JSON spec:
+     *
+     * @items {"type":"string","format":"email"}
+     */
+    emails: string[];
+}
+```
+
+Translation:
+
+```json
+{
+    "$ref": "#/definitions/ShapesData",
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "definitions": {
+        "Shape": {
+            "properties": {
+                "sizes": {
+                    "description": "Specify individual fields in items.",
+                    "items": {
+                        "minimum": 0,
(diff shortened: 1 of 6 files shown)


dobesv:
Potentially fixes https://github.com/YousefED/typescript-json-schema/issues/236

README.md
```diff
@@ -161,6 +161,65 @@ will be translated to
 
 Note that we needed to use `@TJS-type` instead of just `@type` because of an [issue with the typescript compiler](https://github.com/Microsoft/TypeScript/issues/13498).
 
+You can also override the type of array items, either listing each field in its own annotation or one
+annotation with the full JSON of the spec (for special cases).  This replaces the item types that would
```
domoritz:
```suggestion
annotation with the full JSON of the spec (for special cases). This replaces the item types that would
```

test/programs/annotation-items/main.ts
```diff
@@ -0,0 +1,34 @@
+interface MyObject {
+  /**
+   * @items {"type":"integer"}
+   */
+  a: number[];
```
domoritz: Why not use this pattern? 

```ts
{
  a: Integer[];
}

/**
 * @type integer
 */
type Integer = number;
```

dobesv: I haven't gotten that to work, for me it always comes through as just a plain number.

I couldn't find a test case showing that this is possible, and when I added one, the test failed to get the desired result.

dobesv: Has that worked for you?  I wonder if it's something that stops working in TypeScript 3.7

domoritz: It works in https://github.com/vega/ts-json-schema-generator. Anyway, I think it's the right way to support item typings so I'll close this pull request.

dobesv:
I can't use https://github.com/vega/ts-json-schema-generator, it barfs with some parse error or something, and your proposed solution does not work.

What am I to do?

dobesv:
Extremely disappointing 😭

domoritz:
I'm sorry to disappoint you. I think the right thing would be to figure out how to support 

```ts
/**
   * @items {"type":"integer"}
   */
  a: number[];
```

domoritz:
Hmm, thinking about this, this may actually not work since aliases are not fully supported here. I think you're right.

domoritz:
I thought this was a hack to solve something where a better solution would be feasible but I don't think it is.

dobesv:
Tests will not pass in TypeScript 3.7 for this PR until the other PR that fixes support for the `@TJS-` prefixed annotations is merged.

dobesv:
Thank you for reconsidering!

domoritz:
I'm human ;-)

Issue #322, opened 2019-11-08
Typing to integer ignored


lstefanou:
It seems integer typing is broken currently; the following code ('foo.ts')
```
/**
 * @minimum 1
 * @TJS-type integer
 */
export type natural = number;

class Foo {
  bar: natural;
}
```
results in the following JSON schema:
```
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "properties": {
        "bar": {
            "minimum": 1,
            "type": "number"
        }
    },
    "type": "object"
}
```

`bar` should be an integer, but it gets casted to `number`.

We've confirmed this behaviour happening on multiple systems running MacOS, on different versions of this package.

This was working as intended at least 3 months ago (or so), perhaps sooner.

domoritz:
Hmm, I’m not sure I ever overrode the type. Can you check whether there is a test for this kind of behavior and use git bisect to find out when the error was introduced? Thank you.

EelcoHoogendoorn:
@domoritz we tried installing the latest few versions of your package, including ones known to work previously on our machines, but got the same result. I figured perhaps there is an unpinned dependency in your project somewhere that subtly changed its behavior; but insofar as i understand npm packaging (not at all), everything looks pretty pinned.

If there is a test in your current suite that tests exactly this, I havnt found it.

domoritz:
Thank you for checking! I looked for the keyword `integer` and found that we support this:

```ts
interface Foo {
	/**
     * @asType integer
     */
	anInt: number;
}
```

Does this work for you? Can you help update the docs if you think this should be made clearer?

domoritz:
It would be really cool if this library could detect `@multipleOf n` (where `n` is integer) and then automatically convert to type `integer`.

EelcoHoogendoorn:
> We couldn’t find any code matching 'asType' in YousefED/typescript-json-schema

No luck; did you make a typo there?

EelcoHoogendoorn:
> It would be really cool if this library could detect `@multipleOf n` (where `n` is integer) and then automatically convert to type `integer`.

Not saying you are wrong; but that would require the really cool functionality of the plain primitive types working as intended :P.

domoritz:
Ups, my fault. I was looking in https://github.com/vega/ts-json-schema-generator.

lstefanou:
[...]

PR #329, opened 2019-12-05
Add test case for natural number type alias

dobesv:
Test case for https://github.com/YousefED/typescript-json-schema/issues/322


domoritz:
Is this pull request still useful? Since the code has changed and the pull request is no longer mergeable, please send a new one.

PR #330, opened 2019-12-05
Fix processing of @tjs- prefixed jsdoc annotations in newer typescript versions

dobesv:
In newer versions of typescript, the annotation is being taken as a whole instead of being cut at the '-' character.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
 language: node_js
+env:
+  - TYPESCRIPT_VERSION=3.5
+  - TYPESCRIPT_VERSION=3.6
+  - TYPESCRIPT_VERSION=3.7
 node_js:
   - "11"
 script:
-  - npm run lint
-  - npm run test
+  - yarn add typescript@$TYPESCRIPT_VERSION
+  - yarn lint
+  - yarn test
--- a/test/schema.test.ts
+++ b/test/schema.test.ts
@@ -2,7 +2,7 @@ import * as Ajv from "ajv";
 import { assert } from "chai";
 import { readFileSync } from "fs";
 import { resolve } from "path";
-
+import { versionMajorMinor as typescriptVersionMajorMinor } from "typescript";
 import * as TJS from "../typescript-json-schema";
 
 const ajv = new Ajv();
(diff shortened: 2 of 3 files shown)


dobesv:
Original issue report: https://github.com/YousefED/typescript-json-schema/issues/322

dobesv:
@domoritz Is there a way to test with different versions of typescript ?  I think you might need to update the travis-ci configuration somehow.

dobesv:
I've updated the travis config, looks like it might try multiple typescript versions now.

dobesv:
@domoritz I didn't add a new test because the existing tests cover this case.  Just trying to run the tests in typescript 3.7 will fail.  So instead I updated the travis config to run the tests in TypeScript 3.7 and Typescript 3.5.  Does that work for you?

dobesv:
Hmm, I'm not sure what to do with that failing test.  It's not related to my change, I suppose it's just been broken in TypeScript 3.7 all along.  Do you know anything about that test that might help me understand how it should be fixed?  It looks very mysterious to me.

dobesv:
I don't know what `generic-hell` is trying to test, so I disabled it for typescript 3.7+ for now.  Perhaps it can be repaired later.

domoritz:
Oh okay. If the test fails with a previous version, great.

domoritz:
If you want to update typescript, please fix the test. Otherwise, let's just fix to an old version of Typescript.

dobesv:
@domoritz I don't know how to fix the test.  Any ideas?

domoritz:
I don't actively maintain this repo anymore since I have too many other projects to look at. I still review and merge prs, though.

dobesv:
The current situation is that there are a number of tests that fail in TypeScript 3.7 on master.  This PR fixes all of them but one, which is kind of a weird looking test that I don't know what its purpose is.

I think there's probably more value in testing / maintaining compatibility across versions of TypeScript than in having that one very specific test working.

Possible solutions in order of my preference:

1. Leave the PR as-is
2. Delete the part of the test that gives different results between TypeScript versions, so the tests pass in all versions but whatever case that one thing was trying to cover is no longer covered.  Looking at the test I can't figure out who would care about that particular bit of behavior, maybe the test itself is not actually testing anything of use.
3. Only test in TypeScript 3.7 and update that test so it just matches to whatever TypeScript 3.7 produces.  I feel that in this case if nobody depended on the output of that case, the test could have been deleted as it delivers no value.
4. Try to change the tool to produce the same output in TypeScript 3.7 as currently is shown in the test.  This is beyond my level of ability/commitment right now and might even be impossible.

Are any of these solutions acceptable to you?

domoritz:
Since you cleanly disabled the test for one specific version, I think this looks good.

PR #342, opened 2020-03-13
Fix #341

### LINKED ISSUE #341: Type aliases overridden by a schema override are still processed
If a type alias is overridden using `setSchemaOverride()`, the value of the type alias is still processed. It should be ignored.

Example:

```ts
interface All {}

type Some = Partial<All>;

interface MyObject {
  some: Some;
}
```

Expected schema:

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "definitions": {
    "Some": {
      "type": "string"
    }
  },
  "properties": {
    "some": {
      "$ref": "#/definitions/Some"
    }
  }
}
```

Actual schema:

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "definitions": {
    "Partial<All>": {
      "type": "object"
    },
    "Some": {
      "type": "string"
    }
  },
  "properties": {
    "some": {
      "$ref": "#/definitions/Partial<All>"
    }
  }
}
```

remcohaszing:
Please:

### DIFF
--- a/test/programs/type-alias-schema-override/main.ts
+++ b/test/programs/type-alias-schema-override/main.ts
@@ -1,0 +1,7 @@
+interface All {}
+
+type Some = Partial<All>;
+
+interface MyObject {
+  some: Some;
+}
--- a/test/programs/type-alias-schema-override/main.ts
+++ b/test/programs/type-alias-schema-override/main.ts
@@ -3,5 +3,5 @@ interface All {}
 type Some = Partial<All>;
 
 interface MyObject {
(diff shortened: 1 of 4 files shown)


domoritz:
Please remove typescript-json-schema-0.42.0.tgz

remcohaszing:
Oops! Removed it.

I force pushed to prevent a binary from ending up in the git tree.

domoritz:
Thank you

PR #343, opened 2020-03-18
Simplify setup

remcohaszing:
This simplifies the setup around testing and building the project by embracing more standard configurations.

- Mocha is configured using a configuration file, removing the need for CLI options in npm scripts.
- Mocha uses `ts-node/register` instead of `source-map-support/register`. This removes the need to build the project before running tests. This also removes the need for an explicit `source-map-support` dependency.
- The `prepack` script is used for building the project. This makes sure the project is always built before packing or releasing.
- Only relevant files are included in the published package.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -19,17 +19,11 @@
         "noLib": false,
         "preserveConstEnums": true,
         "sourceMap": true,
-        "watch": false,
-        "typeRoots" : ["node_modules/@types"]
+        "watch": false
     },
-    "include": [
-        "**/*.ts",
-        "node_modules/@types/**/*.d.ts"
-    ],
     "exclude": [
-        "node_modules",
         "example",
-        "test/programs",
+        "test",
         "dist"
     ],
     "compileOnSave": true,
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ yarn-error.log
 *~
 .#*
 dist/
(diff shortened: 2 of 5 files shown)


domoritz:
Thank you. I think a build script is still useful if you link the projector. Let’s keep the command.

remcohaszing:
Sorry, I don’t know what you mean by _if you link the projector_.

Anyway, the `prepare` script is an actual script used by npm to run tasks such as building the package. `build` is a custom script name.

domoritz:
Sorry, autocorrect, I wanted to say "link the project" using npm link to generate a schema somewhere else while trying out new code here.

remcohaszing:
I restored the build script. However, one could also run `npm run prepack` / `yarn prepack` instead of `npm run build` / `yarn build`.

tsconfig.json
```diff
@@ -19,17 +19,11 @@
...
     },
-    "include": [
-        "**/*.ts",
-        "node_modules/@types/**/*.d.ts"
-    ],
     "exclude": [
-        "node_modules",
         "example",
-        "test/programs",
+        "test",
```
domoritz: Does this mean we cannot jump to definitions in src from tests or find references in tests from src anymore?

domoritz:
Since the code ha changed and the pull request is no longer mergable, please send a new one.

PR #367, opened 2020-09-08
Added explicit function return types

RohanTalip:
Please:

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -114,7 +114,7 @@ export type SymbolRef = {
   symbol: ts.Symbol;
 };
 
-function extend(target: any, ..._: any[]) {
+function extend(target: any, ..._: any[]): any {
     if (target == null) { // TypeError if undefined or null
       throw new TypeError("Cannot convert undefined or null to object");
     }
@@ -154,7 +154,7 @@ function unique(arr: string[]): string[] {
 /**
  * Try to parse a value and returns the string if it fails.
  */
-function parseValue(value: string) {
+function parseValue(value: string): any {
     try {
         return JSON.parse(value);
     } catch (error) {
@@ -199,7 +199,7 @@ const simpleTypesAllowedProperties = {
     description: true
 };
 
-function addSimpleType(def: Definition, type: string) {
+function addSimpleType(def: Definition, type: string): boolean {
     for (const k in def) {
         if (!simpleTypesAllowedProperties[k]) {
             return false;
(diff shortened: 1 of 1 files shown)


domoritz:
Why does this change improve the code?

RohanTalip:
It helps me at least, as someone new to this code, to see at a glance what type(s) each function is returning, without relying on my IDE (VS Code) to show me that and also without needing to scroll down and find all return statements.

If you think it doesn't add much, feel free to close the pull request.

typescript-json-schema.ts
```diff
@@ -570,7 +570,7 @@ export class JsonSchemaGenerator {
         return undefined;
     }
 
-    private getDefinitionForProperty(prop: ts.Symbol, node: ts.Node) {
+    private getDefinitionForProperty(prop: ts.Symbol, node: ts.Node): Definition|null {
```
domoritz, line 573:
```suggestion
    private getDefinitionForProperty(prop: ts.Symbol, node: ts.Node): Definition | null {
```

domoritz (changes requested):

domoritz:
I think it's helpful. Can you find out why Travis tests are not running?

@YousefED did you disable Travis recently?

typescript-json-schema.ts
RohanTalip: > I think it's helpful.

Great!

> Can you find out why Travis tests are not running?

I'm not familiar with the Travis setup yet ...

By the way, would using a code formatter like Prettier be useful?  (I see that https://github.com/vega/ts-json-schema-generator also uses Prettier)

RohanTalip:
> I think it's helpful.

Great!

> Can you find out why Travis tests are not running?

I'm not familiar with the Travis setup yet ...

By the way, would using a code formatter like Prettier be useful?  (I see that https://github.com/vega/ts-json-schema-generator also uses Prettier)

typescript-json-schema.ts
RohanTalip: FYI, I fixed this by pushing another commit before I realised that I could just click on the `Commit suggestion` button.

domoritz:
Yes, it would be if we can enforce it. Maybe switching to github actions would be a good idea (vega/ts-json-schema-generator uses it as well).

YousefED:
> I think it's helpful. Can you find out why Travis tests are not running?
> 
> @YousefED did you disable Travis recently?

Didn't change anything. Maybe Travis had a hiccup as I think it's running (but failing) now?

PR #377, opened 2020-09-15
OpenAPI format generation

Malien:
[OpenAPI specification data types](https://swagger.io/docs/specification/data-models/data-types/) uses JSON schema to formulate the shape of the data. It is pretty much identical to JSON schema, except it differs from the specification in two ways:

- `null` is not a type in the OpenAPI. Nullability is represented by setting `"nullable": true`
- Type cannot be an array of types, instead `anyOf` or `oneOf` should always be used.

In this PR I've added this compatibility layer behind the `openapiFormat` flag (or `--openapiFormat` in the cli). Added some basic test coverage, and updated README.md a bit.

I've modified this project for my own use to generate (annoying to type otherwise) JSON schema inside of the OpenAPI spec. Yet I've thought bringing this change upstream can be helpful to others.

Since the difference between JSON schema and OpenAPI data type specification is rather small, there is no issue associated and I've just implemented it myself.

_P.S. This is my first OS contribution. Dunno how it should work_

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ Options:
   --rejectDateType      Rejects Date fields in type definitions.                     [boolean] [default: false]
   --id                  Set schema id.                                               [string]  [default: ""]
   --defaultNumberType   Default number type.                                         [choices: "number", "integer"] [default: "number"]
+  --openapiFormat       Generate OpenAPI variation of JSON schema format.            [boolean] [default: false]
 ```
 
 ### Programmatic use
--- a/test/programs/openapi-strict/main.ts
+++ b/test/programs/openapi-strict/main.ts
@@ -1,0 +1,15 @@
+type NullableString = string | null;
+
+interface ComposedType {
+    str: string;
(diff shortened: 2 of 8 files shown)


domoritz:
@Malien thank you for your pull request. I wonder, would you be able to generate OpenAPI from the generates JSON schema? If so, I don't think we need to add Open API support directly to this package.

Malien:
From what I see, yeah, I think I can. I guess doing it in two passes is a tiny bit slower, but since most of the time taken is ts compilation, it is not that meaningful. There are probably some OpenAPI specific annotations available otherwise, but this PR does not include them. 

Also silly me, just found a [package](https://github.com/openapi-contrib/json-schema-to-openapi-schema) to do exactly that. Convert JSON schema to OpenAPI spec.

I guess sorry for wasting time then.

domoritz:
Great. I like if we can separate concerns. 

> I guess sorry for wasting time then.

You learned something and created a resource for others with this issue.

PR #385, opened 2020-11-23
Convert from Travis CI to GitHub actions

### LINKED ISSUE #384: No need to test multiple TypeScript versions
The Travis test environment tests for various TypeScript versions. However, `typescript-json-schema` has a direct dependency on TypeScript. This means it can be safely assumed the TypeScript version in the version range of `package.json` is used.

If this package should have a dependency on `typescript`, the `TYPESCRIPT_VERSION` environment variables should be removed from `.travis.yml` to significantly reduce test time and save resources.

If the package should work with multiple TypeScript versions, `typescript` should be moved to `peerDependencies` using the version range `^3.7 || ^4` (or alternatively `>= 3.7`).

remcohaszing:
The TypeScript version matrix has been removed, as it isn't needed if
TypeScript is a direct dependency.

Closes #384

Please:


domoritz (approved):
Should we add

```
    # We want to run on external PRs, but not on our own internal PRs as they'll be run by the push to the branch.
    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
```

so we don't run all tests twice?

See https://github.com/vega/ts-json-schema-generator/blob/d5c1e179adbc38108134247910b572dc66c3febf/.github/workflows/test.yml#L9

remcohaszing:
I’m not sure. Coming from GitLab CI, I’m not a huge fan of GitHub actions so far. I’m confused why no jobs ran for this PR.

domoritz:
Let's merge and iterate.

domoritz:
I think we just needed to have a workflow file in master to kick things off. Works now. Thanks for the pull request.

Issue #387, opened 2020-11-25
0.44.0 intersection types regression


laat:
`types.ts` given these types:

```
export interface Links {
    readonly self: HrefLink;
    readonly myLink?: HrefLink & TemplatedLink;
}
export interface HrefLink {
  readonly href: string;
}
export interface TemplatedLink {
  readonly templated: true;
}
```

`program.ts` and a that generates schema:
```
import path from 'path';
import util from 'util'
import * as TJS from 'typescript-json-schema';

const program = TJS.getProgramFromFiles([path.join(__dirname, "types.ts")]);
const schema = TJS.generateSchema(program, "Links", {
  noExtraProps: true, // important
});

console.log(util.inspect(schema, { depth: Infinity }));
```

in 0.43.0 gives the expected intersection:
```
{
  type: 'object',
  properties: {
    self: { '$ref': '#/definitions/HrefLink' },
    myLink: {
      additionalProperties: false,
      type: 'object',
      properties: {
        href: { type: 'string' },
        templated: { type: 'boolean', enum: [ true ] }
      }
    }
  },
  additionalProperties: false,
  definitions: {
    HrefLink: {
      type: 'object',
      properties: { href: { type: 'string' } },
      additionalProperties: false
    }
  },
  '$schema': 'http://json-schema.org/draft-07/schema#'
}
```

in 0.44.0 HrefLink properties is missing in `myLink`:
```
{
  type: 'object',
  properties: {
    self: { '$ref': '#/definitions/HrefLink' },
    myLink: {
      additionalProperties: false,
      type: 'object',
      properties: { templated: { type: 'boolean', enum: [ true ] } }
    }
  },
  additionalProperties: false,
  definitions: {
    HrefLink: {
      type: 'object',
      properties: { href: { type: 'string' } },
      additionalProperties: false
    }
  },
  '$schema': 'http://json-schema.org/draft-07/schema#'
}
```

domoritz:
Could that be coming from https://github.com/YousefED/typescript-json-schema/pull/383? cc @darkowic

darkowic:
Not sure... But we do have a similar test here https://github.com/YousefED/typescript-json-schema/blob/master/test/programs/type-intersection/main.ts and everything is passing on CI.

laat:
I bisected the commits, and it was introduced in b69cd3cf87107f4cac71af6dc50d85780506f8ac

darkowic:
[...]

PR #388, opened 2020-11-26
Test type intersection with reuse

laat:
Add tests for https://github.com/YousefED/typescript-json-schema/issues/387


laat:
I have verified that this test passed before b69cd3c. 

I'm sorry that I cannot be of more help.

PR #383, opened 2020-11-21
Handle recursive types #324

darkowic:
Please:
- [x] Make your pull request atomic, fixing one issue at a time unless there are many relevant issues that cannot be decoupled.
- [x] Provide a test case & update the documentation in the `Readme.md`

Fixes #324, #339, #246 -> Hopefully ;) Tested on examples from this repository https://github.com/Guidewire/ts-to-json-schema-tests (The ExampleProps).

Feedback appreciated.

PR #392, opened 2020-12-15
Fix AJV warning, "$ref: keywords ignored in schema"

### LINKED ISSUE #391: Top-level $ref in 0.44.0
When I upgraded from typescript-json-schema 0.43.0 to 0.44.0, I noticed that AJV started giving a new warning when compiling the generated schema:

> $ref: keywords ignored in schema at path "#/properties/data/properties/pcList"

Sample TypeScript types:

```
interface Parent { name?: string; children: Child[]; }
interface Child { name?: string; parent?: Parent; }
```

The generated schema in 0.43.0 has a `definitions.Parent` section and also duplicates the `Parent` type's properties on the root schema. The generated schema in 0.44.0 has a top-level `"$ref": "#/definitions/Parent"` property. If I understand the discussions at https://github.com/ajv-validator/ajv/issues/1121 correctly, the top-level `$ref` object is technically invalid.

If I've misunderstood this, please let me know. Thanks!

joshkel:
The improved handling for recursive schemas that was added in #383 resulted in type-related keywords being added alongside the `$ref` keywords. Such keywords are ignored according to draft-07 of the JSON Schema, so AJV (and possibly other tools) warn about them.

For more information, see https://github.com/ajv-validator/ajv/issues/1121

To address this, I revised the recursive ref logic to only include "annotation" keywords, based on the list here:

http://json-schema.org/understanding-json-schema/reference/generic.html#annotations

(That page also includes `title`, but typescript-json-schema currently doesn't recognize that.)

Note that draft 2019-09 of JSON Schema changes `$ref` to act like a normal keyword, so this change will no longer be necessary in that version of the spec.

Fixes #391.

Please:

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -372,7 +372,7 @@ const annotationKeywords: { [k in keyof typeof validationKeywords]?: true } = {
     examples: true,
     // A JSDoc $ref annotation can appear as a $ref.
     $ref: true
-}
+};
 
 const subDefinitions = {
     items: true,
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -360,6 +360,20 @@ const validationKeywords = {
     id: true
 };
 
+/**
+ * Subset of descriptive, non-type keywords that are permitted alongside a $ref.
+ * Prior to JSON Schema draft 2019-09, $ref is a special keyword that doesn't
+ * permit keywords alongside it, and so AJV may raise warnings if it encounters
+ * any type-related keywords; see https://github.com/ajv-validator/ajv/issues/1121
+ */
+const annotationKeywords: { [k in keyof typeof validationKeywords]?: true } = {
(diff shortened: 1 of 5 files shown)


typescript-json-schema.ts
```diff
@@ -360,6 +360,20 @@ const validationKeywords = {
     id: true
 };
 
+/**
+ * Subset of descriptive, non-type keywords that are permitted alongside a $ref.
+ * Prior to JSON Schema draft 2019-09, $ref is a special keyword that doesn't
+ * permit keywords alongside it, and so AJV may raise warnings if it encounters
+ * any type-related keywords; see https://github.com/ajv-validator/ajv/issues/1121
```
darkowic, line 367: Yes! This is what I should have done when fixing the other issue but it was too much back then. 👍

darkowic (approved):

Issue #394, opened 2020-12-22
Top-level type annotation failed


ExcitedSpider:
Reproduce:

```js
/** This annotation not  be include */
type Shape2 = { kind: string; size: number };
```

```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "properties": {
        "kind": {
            "type": "string"
        },
        "size": {
            "type": "number"
        }
    },
    "type": "object"
}
```

The annotation was given up.

ExcitedSpider:
Hi, I found  something that maybe helpful. I forked the repo and marked.

In  method `JsonSchemaGenerator/getTypeDefinition`:
[commit diff](https://github.com/ExcitedSpider/typescript-json-schema/commit/d3438168e5a0e0b0fe15585f678ea277d9df607e)

domoritz:
Please send a pull request.

ExcitedSpider:
Hi, pull request sent.#395 
Some test case failed due to additional description.

domoritz:
fixed by https://github.com/YousefED/typescript-json-schema/pull/395

PR #395, opened 2020-12-22
fix: top-level alias type comments parse

ExcitedSpider:
Please:

As discussed in #394 , parse comment for top-level type.


domoritz:
Thank you for the pull request. Can you please fix the test cases? You can either correct the specs if that's needed or fix the code.

ExcitedSpider:
OK, I'll do it.

ExcitedSpider:
Hi, I've corrected the test case by adding the parsed comment for `type` alias.

domoritz:
fixes #394

PR #397, opened 2020-12-31
Fix tuple handling (maxItems vs. additionalItems)

### LINKED ISSUE #393: minItems / maxItems
`minItems ` and `maxItems` should be added for tuples according to https://ajv.js.org/docs/strict-mode.html#prevent-unexpected-validation

Otherwise there is this kind of warning in strict mode with ajv:
```
strict mode: "items" is 7-tuple, but minItems or maxItems/additionalItems are not specified or different
```

joshkel:
If I understand correctly, typescript-json-schema's use of additionalItems was incorrect; it had the result of allowing any number of additional items to be added to the end of the tuple, instead of restricting the tuple to only the listed items. It was also causing warnings in AJV 7's new strict mode:
    
> strict mode: "items" is 2-tuple, but minItems or maxItems/additionalItems are not specified or different
    
For more information:
    
- https://github.com/ajv-validator/ajv/blob/v7.0.0/docs/strict-mode.md#prohibit-unconstrained-tuples
- https://json-schema.org/understanding-json-schema/reference/array.html#tuple-validation
    
Fixes #393.

### DIFF
--- a/test/programs/type-aliases-fixed-size-array/schema.json
+++ b/test/programs/type-aliases-fixed-size-array/schema.json
@@ -1,15 +1,5 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "additionalItems": {
-        "anyOf": [
-            {
-                "type": "string"
-            },
-            {
-                "type": "number"
-            }
-        ]
-    },
     "items": [
         {
             "type": "string"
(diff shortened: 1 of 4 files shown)


domoritz:
Thank you for the fix. I made a new release.

PR #399, opened 2021-01-18
Add require functionality in annotations

martin-tichovsky-s2:
Please:

This functionality allows you to enclose a variable (object, array,...) into your property. In my case I needed to have an interface and provide an example in generated JSON schema. But some examples are too big to write them in anotation and we even can't check them by tsc. With 'require' it allows you to import a real object, array, etc. into your provided property. Examples require an array, so when is defined **@examples** or **@TJS-examples** then you must require an array. Every 'require' will fail the JSON schema generation when there will be some errors. This 'require' functionality isn't limited only for the tag 'examples' (but can be easily changed in the code). For big projects and big interfaces is this 'require' functinality really needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typescript-json-schema",
-  "version": "0.48.0",
+  "version": "0.47.0",
   "description": "typescript-json-schema generates JSON Schema files from your Typescript sources",
   "main": "dist/typescript-json-schema.js",
   "typings": "dist/typescript-json-schema.d.ts",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typescript-json-schema",
-  "version": "0.47.0",
+  "version": "0.48.0",
   "description": "typescript-json-schema generates JSON Schema files from your Typescript sources",
   "main": "dist/typescript-json-schema.js",
   "typings": "dist/typescript-json-schema.d.ts",
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "@types/json-schema": "^7.0.6",
     "glob": "^7.1.6",
     "json-stable-stringify": "^1.0.1",
+    "ts-node": "^9.1.1",
     "typescript": "^4.1.3",
     "yargs": "^16.2.0"
   },
@@ -62,7 +63,6 @@
     "mocha": "^8.2.1",
     "prettier": "^2.2.1",
     "source-map-support": "^0.5.19",
-    "ts-node": "^9.1.1",
     "tslint": "^6.1.3"
   },
   "scripts": {
--- a/test/schema.test.ts
+++ b/test/schema.test.ts
@@ -274,8 +274,6 @@ describe("schema", () => {
         });
         assertSchema("annotation-items", "MyObject");
 
-        assertSchema("annotation-required", "MyObject");
-
         assertSchema("typeof-keyword", "MyObject", { typeOfKeyword: true });
 
         assertSchema("user-validation-keywords", "MyObject", {
@@ -459,4 +457,10 @@ describe("tsconfig.json", () => {
             assert.throws(() => generator.getSchemaForSymbol("IncludedOnlyByTsConfig"));
         }
     });
+});
+
+describe("Required", () => {
+    // this part is needed to resolve ts script internaly
+    require("ts-node/register");
+    assertSchema("annotation-required", "MyObject");
 });
--- a/test/schema.test.ts
+++ b/test/schema.test.ts
@@ -274,6 +274,8 @@ describe("schema", () => {
         });
         assertSchema("annotation-items", "MyObject");
 
+        assertSchema("annotation-required", "MyObject");
+
         assertSchema("typeof-keyword", "MyObject", { typeOfKeyword: true });
 
         assertSchema("user-validation-keywords", "MyObject", {
--- a/test/schema.test.ts
+++ b/test/schema.test.ts
@@ -459,8 +459,8 @@ describe("tsconfig.json", () => {
     });
 });
 
-describe("Required", () => {
-    // this part is needed to resolve ts script internaly
-    require("ts-node/register");
-    assertSchema("annotation-required", "MyObject");
+describe("Functionality 'required' in annotation", () => {
+    assertSchema("annotation-required", "MyObject", {
+        tsNodeRegister: true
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "typescript-json-schema",
-  "version": "0.47.0",
+  "version": "0.48.0",
```
domoritz, line 3: Please don't bump the version here

typescript-json-schema.ts
```diff
@@ -7,11 +7,13 @@ import { JSONSchema7 } from "json-schema";
 export { Program, CompilerOptions, Symbol } from "typescript";
 
 const vm = require("vm");
+require("ts-node/register");
```
domoritz, line 10: I don't think this should be here.

typescript-json-schema.ts
```diff
@@ -7,11 +7,13 @@ import { JSONSchema7 } from "json-schema";
...
 
 const REGEX_FILE_NAME_OR_SPACE = /(\bimport\(".*?"\)|".*?")\.| /g;
 const REGEX_TSCONFIG_NAME = /^.*\.json$/;
 const REGEX_TJS_JSDOC = /^-([\w]+)\s+(\S|\S[\s\S]*\S)\s*$/g;
 const REGEX_GROUP_JSDOC = /^[.]?([\w]+)\s+(\S|\S[\s\S]*\S)\s*$/g;
+const REGEX_REQUIRE = /^(\s+)?require\((\'@?[a-zA-Z0-9.\/_-]+\'|\"@?[a-zA-Z0-9.\/_-]+\")\)(\.([a-zA-Z0-9_$]+))?(\s+|$)/;
```
domoritz, line 16: I think this needs a comment at least. Maybe it would be better to not write the full regex but instead only define the start of the match. 

Also, needs tests.

domoritz (changes requested):
Thank you for the pull request. Can you just confirm that require is standard functionality in jsdocs?

package.json
martin-tichovsky-s2: Ok, removed.

typescript-json-schema.ts
martin-tichovsky-s2: This part is really necessary because of require typescript files. Without this part isn't ts-node registered and can't resolve typescript files.

typescript-json-schema.ts
martin-tichovsky-s2: I've provided a comment and a large test. Other way I don't know how to split the regex to be more understandable.

domoritz: Awesome. Thank you.

martin-tichovsky-s2:
I don't think a 'require' is a valid value of any jsdocs' tags. As they don't have @examples. But from my point of view is really useful to have such functionality here because as I mention especially in examples is sometimes need to put big objects. And also can be extended for others tags.

typescript-json-schema.ts
domoritz: I don't think that's true. It worked before, right?

domoritz:
Can you give an example beyond examples then?

I will not yet merge this pull request and give it a few days for people to chime in and say they support or don't want this change since it's non-standard.

typescript-json-schema.ts
martin-tichovsky-s2: Yes, but not when you try to require a typescript file. You could require only JS file. An in this case people will want to require (or import) file from they own project, such an example from its own repo. And in that case without ts-node won't work. You can try by yourself, remove this part and run tests. You will get an error that module can't be found.

domoritz: You can either import the js files we compile and put on npm or call the ts file with ts node. This is how it always worked and I don't want to change it.

martin-tichovsky-s2:
When I use @examples tag in comment it will work yes? Because it's the same as @TJS-examples. Jsdocs doesn't have @examples tag too so I would not like to talk about standards. Yes, this is not standard in jsdocs. But I assume some tags used here in this repo are not standards too. So from this point I supposed that use identical syntax of 'require' will be useful.

domoritz:
Agreed. I will merge after giving others a chance to voice their thoughts (if there are any).

typescript-json-schema.ts
domoritz: If it's needed for the require calls, you should figure out a way to only activate ts node for those calls.

martin-tichovsky-s2: Aha ok, I think we both are looking on this problem different way. You suppose that this package will be used somewhere live, maybe on some web page? But I assume and for our purpose is need to use the "typescript-json-schema" only internally for generating json schemas before deploy. So from our side we don't have compiled scripts when is json generation called. Therefore I used this possibility to load typescript files. But I founded that in this case we need to require "ts-node/register" before using 'generateSchema'. So ok, I removed this part and for people who want to use this functionality internally, must use require("ts-node/register") before using "typescript-json-schema". I updated tests and put the test at the end of tests.

test/schema.test.ts
```diff
@@ -460,3 +458,9 @@ describe("tsconfig.json", () => {
         }
     });
 });
+
+describe("Required", () => {
+    // this part is needed to resolve ts script internaly
+    require("ts-node/register");
```
domoritz, line 464: Hmm, does this mean I cannot call `typescript-json-schema` with just node (for example with npx)?

typescript-json-schema.ts
domoritz: I don't think that's an option as I explained below. Maybe we need to add this back but you also need to make ts-node a dependency then (which I find a bit worrisome since it's a complex dependency).

martin-tichovsky-s2: It's true, so I added tsNodeRegister option to enable ts-node/register. This will works with command such as "typescript-json-schema src/generated-type.ts ShapesData --tsNodeRegister". Or it's possible to set it in settings when calling 'generateSchema()'. Maybe this will be better option then making ts-node as dependency, because in production, nobody will have ts scripts.

domoritz: Good idea.

martin-tichovsky-s2: You have absolutely right, I make ts-node as dependency and now it's working correctly. Unfortunately I haven't found any better way.

domoritz:
Thank you @martin-tichovsky-s2

PR #405, opened 2021-02-23
Comments and utility types

marneborn:
Please:

Currently (since PR #395 ) types created with utility types get the description on the type in lib.es5.d.ts.
For example:
```
/**
 * This comment is ignored. Should it be included?
 */
export type MyObject = Pick<BigThing, "prop1">;
interface BigThing {
  prop1: string;
  prop2: string;
};
``` 

Results in:
```
{
  "description": "From T, pick a set of properties whose keys are in the union K",
  "type": "object",
  "properties": {
    "prop1": {
      "type": "string"
    }
  },
  "required": [
    "prop1"
  ],
  "$schema": "http://json-schema.org/draft-07/schema#"
}
 ```
This is the definition from lib.es5.d.ts
```
/**
 * From T, pick a set of properties whose keys are in the union K
 */
type Pick<T, K extends keyof T> = {
    [P in K]: T[P];
};
```

This PR results in:
```
{
  "type": "object",
  "properties": {
    "prop1": {
      "type": "string"
    }
  },
  "required": [
    "prop1"
  ],
  "$schema": "http://json-schema.org/draft-07/schema#"
}
 ```

This matches (I think) the functionality before PR #395 for this case. Should I try to get this comment into the description?
Ie, should the this test give this instead? 
```
{
  "description": "This comment is ignored. Should it be included?",
  "type": "object",
  "properties": {
    "prop1": {
      "type": "string"
    }
  },
  "required": [
    "prop1"
  ],
  "$schema": "http://json-schema.org/draft-07/schema#"
}
 ```

### DIFF
--- a/test/programs/comments-from-lib/main.ts
+++ b/test/programs/comments-from-lib/main.ts
@@ -1,10 +1,10 @@
 /**
- * Want this comment, but doesn't work at the moment
+ * Use this comment
  */
 export type MyObject = Pick<BigThing, "prop1">;
 
 /**
- * This comment should be ignored
+ * Not this comment though
  */
 interface BigThing {
   prop1: string;
--- a/test/programs/comments-from-lib/main.ts
+++ b/test/programs/comments-from-lib/main.ts
@@ -3,6 +3,9 @@
  */
 export type MyObject = Pick<BigThing, "prop1">;
 
+/**
+ * Not this comment though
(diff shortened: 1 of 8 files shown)


domoritz (changes requested):
Yes, I think the comment should be included (as in your test case).

Issue #408, opened 2021-03-09
Add an option to create output directory if necessary


arnaudcourtecuisse:
Hi, 

I use this tool to generate files in a directory that is not versioned, and only used for schemas: `./dist/schemas`
At the moment the script throws an error:
```
Error: Unable to write output file: ENOENT: no such file or directory, open './dist/schemas/payload.json'
    at ***/poc/node_modules/typescript-json-schema/typescript-json-schema.ts:1668:23
    at fs.js:1448:7
    at FSReqCallback.oncomplete (fs.js:171:23)
```

Would it be possible to add an option (maybe `--createParentDir`) that automatically creates the output directory if it does not exist ?

domoritz:
Happy to accept a pull request.

PR #409, opened 2021-03-10
Create output parent directory when necessary

arnaudcourtecuisse:
Implement #408 with tests

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -1678,11 +1678,17 @@ export async function exec(filePattern: string, fullTypeName: string, args = get
     const json = stringify(definition, { space: 4 }) + "\n\n";
     if (args.out) {
         return new Promise((resolve, reject) => {
-            require("fs").writeFile(args.out, json, function (err: Error) {
-                if (err) {
-                    return reject(new Error("Unable to write output file: " + err.message));
+            const fs = require("fs");
+            fs.mkdir(path.dirname(args.out), { recursive: true }, function (mkErr: Error) {
+                if (mkErr) {
+                    return reject(new Error("Unable to create parent directory for output file: " + mkErr.message));
                 }
-                resolve();
+                fs.writeFile(args.out, json, function(wrErr: Error) {
+                    if (wrErr) {
+                        return reject(new Error("Unable to write output file: " + wrErr.message));
+                    }
+                    resolve();
+                });
             });
         });
     } else {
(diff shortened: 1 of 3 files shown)


domoritz:
I think this should be the default. Let's not add an option.

domoritz (changes requested):
I think this should be the default. Let's not add an option.

typescript-json-schema.ts
```diff
@@ -1677,12 +1688,28 @@ export function exec(filePattern: string, fullTypeName: string, args = getDefaul
 
     const json = stringify(definition, { space: 4 }) + "\n\n";
     if (args.out) {
-        require("fs").writeFile(args.out, json, function (err: Error) {
-            if (err) {
-                throw new Error("Unable to write output file: " + err.message);
+        const fs = require("fs");
+        fs.writeFile(args.out, json, function (err: Error) {
```
domoritz, line 1692: Let's simplify this code and always run `fs.mkdir`. It should be no-op when the dir already exists.

typescript-json-schema.ts
```diff
@@ -1653,7 +1653,18 @@ function normalizeFileName(fn: string): string {
...
+    if (err !== null) {
+        throw err;
+    }
+}
+
+export function exec(
+    filePattern: string,
+    fullTypeName: string,
+    args = getDefaultArgs(),
+    callback: (err: Error | null) => void = defaultCallback
```
domoritz, line 1666: Let's not add a callback. Just throw an error like before.

domoritz (changes requested):

typescript-json-schema.ts
domoritz: If you want, you can mark the function as async in a follow up pull request. I prefer promises over callbacks.

arnaudcourtecuisse: The callback is necessary to trigger the tests in `test/out-option.test.ts`. The only alternative I see would be to use `fs.writeFileSync` and `fs.mkdirSync`. Do you want me to do so ? I'm open to suggestions if you have other ideas.

domoritz: I see. Yes, let's use the sync versions for now.

arnaudcourtecuisse: I prefer promises to. I useda callback to minimize the change.

Let's put this PR on hold. I'll open another to use a promise, and once the promise is merged, I'll update this one.

domoritz: Sounds good. Thanks!

arnaudcourtecuisse:
I rebased my branch and used a promise instead of a callback

PR #416, opened 2021-04-13
Cache the generator to improve performance

murbanowicz:
This is a simple change, but in our case, it helped to reduce the tests time by ~95%.
This PR makes the generator being reused between `generateSchema` calls.

I do not use this library anywhere else than some tests in our code, so I may be missing other use-cases. If that is the case, please help me make it work for cases that you know serve everyone.

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -1589,13 +1589,18 @@ export function buildGenerator(
     }
 }
 
+let generator: JsonSchemaGenerator | null | undefined;
+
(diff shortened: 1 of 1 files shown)


domoritz:
Thanks for the pull request. I don't think the caching is correct since arguments can change. 

Also, you can just reuse a generator you create in your app so I'd say that's a more explicit and clean approach.

murbanowicz:
@domoritz how can I reuse the generator then?

Issue #415, opened 2021-04-13
Reuse generator?


murbanowicz:
Hi,

Would you accept simple PR which would cache and reuse the generator during the same session?
Setting up the generator take few seconds, so in our case, when we use `generateSchema` over 100 times in test, it takes crazy long time, while with a 'hacked' solution, it took few seconds.

What I mean is doing something like this:
```typescript
let generator;
function generateSchema(program, fullTypeName, args, onlyIncludeFiles) {
    if (generator=== undefined) {
        generator= buildGenerator(program, args, onlyIncludeFiles);
    }
```

domoritz:
Definitely!

murbanowicz:
Cool. Will do that right now.
In our case, this "hack" decreased the test time from over 10m to about 15s.

MarcusRiemer:
For the record: I had a similar issue and just "solved" the problem by generating a `Makefile` that could be called with `make -j` to launch as many processes as the system could meaningfully execute. A proper solution would be greatly appreciated.

murbanowicz:
I will start simple PR and we can discuss/work together on that. 
I am fairly new here, but this issue was hurting our CI times so much, so I had to fix it in some way.

murbanowicz:
Done: #416 - let's discuss it there.

domoritz:
You could use the api to create a single generator. I don't think we can easily cache in the general case since inputs may change.

murbanowicz:
I don't think I can or at least I cannot find the way.

We could cache based on params with a simple map <hash of params, generator> of generators maybe?

domoritz:
```js
const generator = buildGenerator(program, args, onlyIncludeFiles);
```

Now you can reuse the `generator` variable for all your tests.

murbanowicz:
Agree to some degree, but `generateSchema` does some other stuff that should stay there and not be repeated elsewhere.
Take a look at provided changes. It caches if it can with respect to changing input.

domoritz:
How about we refactor the code to add a function that takes in a generator then? This way we allow composition but don't need to implement caching?

murbanowicz:
That could work.
We can then accept optional generator as optional parameter but when it is not passed, it is still being created. Then, the caching is on user side, while there is no breaking change.

domoritz:
That also sounds like a good option. Let's do that.

murbanowicz:
Ok. Will do that in the evening or tomorrow.

[...]

PR #418, opened 2021-04-24
Allow to reuse generator in generateSchema

murbanowicz:
I have run `yarn style` to stick to the repo style and it looks like some files where changed without that, so that is why changes are also affecting other files.

Fixes https://github.com/YousefED/typescript-json-schema/issues/415


domoritz:
Thank you

PR #436, opened 2021-09-20
Clear reffered symbols before generating schemas.

ejoebstl:
This allows to re-use an existing `generator` instance when using typescript-json-schema programatically.

Otherwise, the referred definitions continue to accumulate, leading to larger and larger schemas with unnecessary definitions.

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -1385,6 +1385,9 @@ export class JsonSchemaGenerator {
         if (!this.allSymbols[symbolName]) {
             throw new Error(`type ${symbolName} not found`);
         }
+
+        this.reffedDefinitions = { }
+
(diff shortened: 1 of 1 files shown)


domoritz:
I'll make this pull request as draft. Please mark it as ready when the tests pass.

ejoebstl:
Interesting that they fail. I did not expect that.

Do you think this is a breaking change? Or are the tests incorrect?

domoritz:
You need to look into it. If you think the tests are wrong, change them in this pull request.

alfaproject:
Oh man, just noticed this as well. @ejoebstl do you plan on following this up?

domoritz:
Superseded by https://github.com/YousefED/typescript-json-schema/pull/466

PR #449, opened 2021-12-11
replaced 'prepublishOnly' script with 'prepare'

random42:
- replaced `prepublishOnly` script with `prepare` (here's [why](https://stackoverflow.com/a/57829251/7392169))
- using [npm-run-all](https://www.npmjs.com/package/npm-run-all) to run scripts (for readibility and compatibility with both npm and yarn)

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,19 +61,21 @@
     "ajv-formats": "^2.1.1",
     "chai": "^4.3.4",
     "mocha": "^9.1.1",
+    "npm-run-all": "^4.1.5",
     "prettier": "^2.4.1",
     "source-map-support": "^0.5.20",
     "tslint": "^6.1.3"
   },
   "scripts": {
-    "prepublishOnly": "yarn build",
-    "test": "yarn build && mocha -t 5000 --require source-map-support/register dist/test",
+    "prepare": "run-s build",
+    "test": "run-s build test:mocha",
+    "test:mocha": "mocha -t 5000 --require source-map-support/register dist/test",
     "debug": "node --inspect=19248 --inspect-brk -r ts-node/register typescript-json-schema-cli.ts",
     "docs": "./update-docs.js",
     "run": "ts-node typescript-json-schema-cli.ts",
     "build": "tsc -p .",
     "lint": "tslint --project tsconfig.json -c tslint.json --exclude '**/*.d.ts'",
-    "style": "yarn prettier --write *.js *.ts test/*.ts",
+    "style": "prettier --write *.js *.ts test/*.ts",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -61,19 +61,21 @@
...
+    "prepare": "run-s build",
+    "test": "run-s build test:mocha",
+    "test:mocha": "mocha -t 5000 --require source-map-support/register dist/test",
     "debug": "node --inspect=19248 --inspect-brk -r ts-node/register typescript-json-schema-cli.ts",
     "docs": "./update-docs.js",
     "run": "ts-node typescript-json-schema-cli.ts",
     "build": "tsc -p .",
     "lint": "tslint --project tsconfig.json -c tslint.json --exclude '**/*.d.ts'",
-    "style": "yarn prettier --write *.js *.ts test/*.ts",
+    "style": "prettier --write *.js *.ts test/*.ts",
```
domoritz, line 78: Doesn't this mean prettier has to be global?

package.json
```diff
@@ -61,19 +61,21 @@
...
+    "npm-run-all": "^4.1.5",
     "prettier": "^2.4.1",
     "source-map-support": "^0.5.20",
     "tslint": "^6.1.3"
   },
   "scripts": {
-    "prepublishOnly": "yarn build",
-    "test": "yarn build && mocha -t 5000 --require source-map-support/register dist/test",
+    "prepare": "run-s build",
+    "test": "run-s build test:mocha",
```
domoritz, line 71: I don't think we need npm compat and I'd rather not add the dependency.

domoritz (changes requested):

package.json
random42: No, it is installed as a dev dependency, it will just work.

package.json
random42: This dependency is widely used and only as dev, so it does not cost anything. Using the `prepare` script and run-s instead of yarn means that github forks installs will work and especially if users use npm (most do).

random42: And also using `prepare` is what npm [suggests](https://docs.npmjs.com/cli/v8/using-npm/scripts#prepare-and-prepublish).

domoritz: Since we use yarn.lock, I don't want developers to use npm. Using prepare sounds good to me.

random42: I'm going to open a new pull request without the dependency

domoritz: Can you update this pull request instead?

random42: I'm sorry I did not know how to do so, I deleted this branch and created a new pr

PR #456, opened 2022-01-11
fix(deps): update dependency typescript to ~4.5.0

richardkazuomiller:
Please:

I've tested this in my project and it seems to be working 😄


domoritz:
Thank you

PR #461, opened 2022-01-14
Removed invalid tokens from generic type names

### LINKED ISSUE #454: v0.52.0 Unique Names causes generation to fail when using generics
With the 0.5x versions of this package, the following will fail to compile with the `--uniqueNames` option:

```typescript
export interface IBulkOperationEntity<EntityIdentifier, BulkOptions extends object = undefined> {
	entity: EntityIdentifier;
	opts?: BulkOptions;
}

export interface IBulkOperation<EntityIdentifier, BulkOptions extends object> {
	options: BulkOptions;
	entities: IBulkOperationEntity<EntityIdentifier>[];
}

export interface UserBulkBetaInvite extends IBulkOperation<string, {}> {
}
```

Run `typescript-json-schema "src/test.ts" UserBulkBetaInvite --uniqueNames` and you receive the following:
```
PS D:\tsj-repro> npx typescript-json-schema "src/test.ts" UserBulkBetaInvite --uniqueNames
(node:6884) UnhandledPromiseRejectionWarning: Error: Symbol "__type" has no valueDeclaration and 0 declarations.
    at getCanonicalDeclaration (D:\tsj-repro\node_modules\typescript-json-schema\dist\typescript-json-schema.js:258:11)
    at getSourceFile (D:\tsj-repro\node_modules\typescript-json-schema\dist\typescript-json-schema.js:261:23)
    at JsonSchemaGenerator.getTypeDefinition (D:\tsj-repro\node_modules\typescript-json-schema\dist\typescript-json-schema.js:931:34)
    at JsonSchemaGenerator.getDefinitionForProperty (D:\tsj-repro\node_modules\typescript-json-schema\dist\typescript-json-schema.js:510:31)
    at D:\tsj-repro\node_modules\typescript-json-schema\dist\typescript-json-schema.js:805:37
    at Array.reduce (<anonymous>)
    at JsonSchemaGenerator.getClassDefinition (D:\tsj-repro\node_modules\typescript-json-schema\dist\typescript-json-schema.js:803:45)
    at JsonSchemaGenerator.getTypeDefinition (D:\tsj-repro\node_modules\typescript-json-schema\dist\typescript-json-schema.js:1025:26)
    at JsonSchemaGenerator.getSchemaForSymbol (D:\tsj-repro\node_modules\typescript-json-schema\dist\typescript-json-schema.js:1056:24)
    at generateSchema (D:\tsj-repro\node_modules\typescript-json-schema\dist\typescript-json-schema.js:1236:30)
(node:6884) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:6884) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

However, if you leave off `--uniqueNames` then you get the following invalid output:
```
PS D:\tsj-repro> npx typescript-json-schema "src/test.ts" UserBulkBetaInvite  
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "definitions": {
            "properties": {
                "entity": {
                    "type": "string"
                },
                "opts": {
                    "type": "undefined"
                }
            },
            "type": "object"
        }
    },
    "properties": {
        "entities": {
            "items": {
                "$ref": "#/definitions/IBulkOperationEntity<string,undefined>"
            },
            "type": "array"
        },
        "options": {
            "properties": {},
            "type": "object"
        }
    },
    "type": "object"
}
```

rkesters:
Please:

Fixes #454 

Generics type names were being created with '<,>' characters when using theses as URI for the $ref fields they are invalid.  Changing this also fixed #454 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
   },
   "devDependencies": {
     "@types/chai": "^4.2.21",
+    "@types/lodash": "^4.14.178",
     "@types/glob": "^7.1.4",
     "@types/mocha": "^9.0.0",
     "ajv": "^8.6.3",
@@ -64,7 +65,9 @@
     "mocha": "^9.1.3",
     "prettier": "^2.4.1",
     "source-map-support": "^0.5.20",
-    "tslint": "^6.1.3"
+    "tslint": "^6.1.3",
+    "jsonpath-plus": "^6.0.1",
+    "lodash": "^4.17.21"
   },
   "scripts": {
     "prepare": "tsc",
--- a/test/programs/generic-anonymous/schema.json
+++ b/test/programs/generic-anonymous/schema.json
@@ -1,7 +1,7 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
     "definitions": {
-        "MyGeneric<number,string>": {
+        "MyGeneric_number_string_": {
             "properties": {
                 "a": {
                     "type": "number"
@@ -16,7 +16,7 @@
             ],
             "type": "object"
         },
-        "MyGeneric<string,number>": {
+        "MyGeneric_string_number_": {
             "properties": {
                 "a": {
                     "type": "string"
@@ -34,10 +34,10 @@
     },
     "properties": {
         "value1": {
-            "$ref": "#/definitions/MyGeneric<string,number>"
+            "$ref": "#/definitions/MyGeneric_string_number_"
         },
         "value2": {
-            "$ref": "#/definitions/MyGeneric<number,string>"
+            "$ref": "#/definitions/MyGeneric_number_string_"
         }
     },
     "required": [
--- a/test/programs/generic-hell/schema.json
+++ b/test/programs/generic-hell/schema.json
@@ -1,7 +1,7 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
     "definitions": {
-        "GenericA<\"alias\">": {
+        "GenericA_\"alias\"_": {
             "properties": {
                 "a": {
                     "enum": [
@@ -15,7 +15,7 @@
             ],
             "type": "object"
         },
-        "GenericA<1>": {
+        "GenericA_1_": {
             "properties": {
                 "a": {
                     "enum": [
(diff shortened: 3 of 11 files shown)


domoritz:
I don't want to make this major change. Please make a fork if you want this change.

CaselIT:
@domoritz would you accept a PR that just url-escapes the url generated by the refs?
Meaning this line https://github.com/YousefED/typescript-json-schema/blob/75e599aaef6a57e849dbb3e1ff0b6f3749937b10/typescript-json-schema.ts#L1417

would become 
```ts
$ref: `${this.args.id}#/definitions/` + encodeURIComponent(fullTypeName),
```

domoritz:
How would that affect what genetics look like? If the json schema spec says something about what's allowed, I'm happy to change things. But I want to make sure things remain readable.

CaselIT:
the generics would remain the same in the definition, so `Foo<string>`, just the `$ref` to them would change to `#/definitions/Foo%3Cstring%3E`, since these are url and should ideally be url escaped.

Some json-schema library are strict about the fact that `$ref` are url, and error if they see an url with angle brackets since that's not a valid character for an url

domoritz:
Having said that, I don't feel too strongly about the change you proposed and would be happy to accept it.

CaselIT:
> I have not seen a library that doesn't think these urls are valid. Would be libraries be open to change?

I'm having issues with this library, that triggered this discussion. https://github.com/Stranger6667/jsonschema/blob/master/crates/jsonschema-py/CHANGELOG.md#changed-5

While annoying I feel like they are doing the correct thing here, since json schema defines $refs as URI, and uri do require percent encoding.

> Having said that, I don't feel too strongly about the change you proposed and would be happy to accept it.

Ok, will do. Do you also prefer I open an issue to reference it in changelogs etc?

domoritz:
Yes, issue and pull request would be great.

DavideCanton:
I could work on this in these days. I'll open a new issue soon.

DavideCanton:
Created #617

PR #467, opened 2022-02-10
docs: auto-update docs on release

puppybits:
1. Docs should automatically get updated on each release via `yarn prepare`. 
2. Also added comment at the top of `api.md` to inform the reader it's an autogenerated file.

Please:


puppybits:
@YousefED Thanks for the great library! I made the script that autogenerates documentation run on `prepare`. That should force it to be updated on every deploy.

PR #482, opened 2022-05-23
Union types may be considered as reffedTypes

NPellet:
Subtypes of unions should be considered eligible for ```asTypeAliasRef``` in ```getTypeDefinition```

e.g.
```
type A = {
  "prop": string
}

type B = {
  "prop": string,
  "refToA"?: A | null 
}
```

should return a reference to A when compiled with ```--aliasRefs```

PS: Sorry for the changes introduced by prettier...


NPellet:
Sorry, screwed up my last commit.

Coming up

Issue #483, opened 2022-05-26
Cannot find module 'tslib' @0.53.1


SimonLdj:
Hi, 
since version 0.53.1 any command fails with `Cannot find module 'tslib'` error
checking `npx typescript-json-schema@0.53.0` working fine.

full stack trace: 
```
~ >npx typescript-json-schema@0.53.1                                                                                                                                                                            [1]
Need to install the following packages:
  typescript-json-schema@0.53.1
Ok to proceed? (y) y
internal/modules/cjs/loader.js:905
  throw err;
  ^

Error: Cannot find module 'tslib'
Require stack:
- /Users/username/.npm/_npx/fc767c80b0695a60/node_modules/path-equal/cjs/index.js
- /Users/username/.npm/_npx/fc767c80b0695a60/node_modules/typescript-json-schema/dist/typescript-json-schema.js
- /Users/username/.npm/_npx/fc767c80b0695a60/node_modules/typescript-json-schema/dist/typescript-json-schema-cli.js
- /Users/username/.npm/_npx/fc767c80b0695a60/node_modules/typescript-json-schema/bin/typescript-json-schema
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:93:18)
    at Object.<anonymous> (/Users/username/.npm/_npx/fc767c80b0695a60/node_modules/path-equal/cjs/index.js:3:15)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12)
    at Module.require (internal/modules/cjs/loader.js:974:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/username/.npm/_npx/fc767c80b0695a60/node_modules/path-equal/cjs/index.js',
    '/Users/username/.npm/_npx/fc767c80b0695a60/node_modules/typescript-json-schema/dist/typescript-json-schema.js',
    '/Users/username/.npm/_npx/fc767c80b0695a60/node_modules/typescript-json-schema/dist/typescript-json-schema-cli.js',
[...]

PR #485, opened 2022-06-01
Fix crashing on tslib not being installed (#483)

BalaM314:
The best commits are those that fix a crash by adding one character.
Oops.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "@types/json-schema": "^7.0.9",
     "@types/node": "^16.9.2",
     "glob": "^7.1.7",
-    "path-equal": "1.1.2",
+    "path-equal": "^1.1.2",
     "safe-stable-stringify": "^2.2.0",
     "ts-node": "^10.2.1",
     "typescript": "~4.6.0",


domoritz:
Please fix the test run.

BalaM314:
It's because the dependency path-equal was updated, so yarn needs to update the yarn.lock file to specify a different version. Should I edit it manually?

domoritz:
If you run yarn in the directory, it will update the lock file.

azarashi2931:
@BalaM314 @domoritz Is this PR working? I am also having trouble with this issue and if it is not working I can take it over.

BalaM314:
The test is failing because I didn't update the lockfile. I just need to clone this repo and run yarn. Forgot this existed, will do in a few hours.

BalaM314:
image
The issue seems to be with this command in the package.json
```
  "scripts": {
    "prepare": "tsc; yarn docs",
```
Changed it to `tsc && yarn docs` and it works.

domoritz:
Thanks. Please update the pull request.

Issue #492, opened 2022-08-06
Not following JS spec


qwelias:
Support `toJSON` method on objects is required for [accurate JS/TS to JSON translation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#description).
`toJSON` can define custom representation of any JS object wich may not match the object itself.
If `toJSON` is not respected by JSON schema generators then generated schema may deviate from actual JSON representation of a JS object.

Example: https://github.com/mongodb/js-bson/blob/main/src/objectid.ts#L202

PR #493, opened 2022-08-09
feat: allow overrides with JSDoc

qwelias:
Please:

Helps workadound https://github.com/YousefED/typescript-json-schema/issues/492 and https://github.com/Coobaha/typed-fastify/issues/56

### DIFF
--- a/test/programs/override/main.ts
+++ b/test/programs/override/main.ts
@@ -1,0 +1,9 @@
+import type { ObjectId } from './third-party'
+
+export type MyObject = {
+  /**
+   * @TJS-type string
+   * @override
+   */
+  _id: ObjectId
+}
--- a/README.md
+++ b/README.md
@@ -228,6 +228,8 @@ Translation:
 
 This same syntax can be used for `contains` and `additionalProperties`.
 
+To completely override types aliased from third-party libraries use `@override`.
+
 ### `integer` type alias
 
 If you create a type alias `integer` for `number` it will be mapped to the `integer` type in the generated JSON schema.
--- a/test/programs/override/schema.json
+++ b/test/programs/override/schema.json
@@ -1,0 +1,13 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "properties": {
+        "_id": {
+            "type": "string"
+        }
+    },
+    "required": [
+        "_id"
+    ],
+    "type": "object"
+}
+
--- a/test/programs/override/third-party.ts
+++ b/test/programs/override/third-party.ts
@@ -1,0 +1,2 @@
+// cannot modify with JSDoc because third-party sources
+export class ObjectId {}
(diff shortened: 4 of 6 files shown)


test/programs/override/main.ts
```diff
@@ -0,0 +1,9 @@
+import type { ObjectId } from './third-party'
+
+export type MyObject = {
+  /**
+   * @TJS-type string
+   * @override
```
domoritz, line 6: Remind me, what happens if we just set `@TJS-type` without `@override`?

qwelias: In this case nothing, it'll just `$ref` to `ObjectId`

domoritz: I see. Would it make sense to make override the default then or not even have an option?

qwelias: What do you mean by `override the default`?
It's not possible to override whatever is coming from third-party lib.
And I didn't want to affect existing behavior, so used `@override`

domoritz: Is the existing behavior reasonable? I'm happy to change it if not.

qwelias: I think we'll converge at least two use-cases if this would be the default one, which are:
1. amending type or adding details (min/max int, etc) (current, doesn't work for aliases)
1. completely overriding a type (proposed in the PR)

I think 1st use-case is completely valid and should be preserved and then fixed (maybe, dunno how yet) at some point.
2nd can exist regardless of whether 1st works for aliases or not.

domoritz: I'm still not getting what @override does. Cans you explain what exactly happens if you don't set it and whether that is only true for third part libraries? Also, please make sure that the docs are clear. 

What I proposed was to not require the additional option and to always override if `@TJS-type` is set.

qwelias: To see exactly what happens you can run the added test case without `@override` comment.
In-short: compiled definition will have `$ref` because the property type is aliased from another place and any other properties defined by JSdoc, but other properties are ignored [because](https://json-schema.org/understanding-json-schema/structuring.html#ref):
> When evaluating a $ref, an implementation uses the resolved identifier to retrieve the referenced schema and applies that schema to the instance.

That is true for any aliased type, sorry for confusion.
___
But former wouldn't matter if I got this right:

> What I proposed was to not require the additional option and to always override if `@TJS-type` is set.

I think I see now, because:
- if it's not aliased, then we know only the `type` part of definition, so any JSDoc is either adding to definition or overrides the `type`, so it's safe to just use whole parsed JSDoc if `@TJS-type` is present
- if it is aliased then any JSDoc indicates the intention for override, but `type` is required on schema, so if user provided `@TJS-type` we can as well safely use parsed JSDoc

Did I get it right?

domoritz: That sounds right. Let's try it out and see whether it works in practice.

qwelias:
in favor of https://github.com/YousefED/typescript-json-schema/pull/493

PR #496, opened 2022-08-17
fix: override alias with prop's JSDoc if type is specified

qwelias:
Please:


qwelias:
@domoritz can you have a look? I'm also working another thing to help avoid having weird generated property names

PR #504, opened 2022-10-10
Added function parameters definition in schema when using `typeOf` keyword

Tiberiu02:
When using the  `typeOf` keyword, type definitions of all functions will now have a "parameters" field containing a list of the type definitions of all function parameters.

This is very useful for fully automated runtime type checking in RPC architectures.

### Example:

Code:
(Note the use of various function syntaxes)

```
function sum(a: number, b: number) {
  return a + b;
}

function prod(a: number, b: number) {
  return a + b;
}

const length = (s: string) => s.length;

const extractAttribute = (obj: object, attribute: string) => obj[attribute];

class Arrays {
  static get(array: number[], index: number) {
    return array[index];
  }
}

export const API = {
  numbers: {
    sum,
    prod
  },
  objects: {
    length,
    extractAttribute,
    extractArrayEl: Arrays.get
  }
};
```

Produced schema:
(note the parameters lists)

```
{
  "type": "object",
  "properties": {
    "api": {
      "type": "object",
      "properties": {
        "numbers": {
          "type": "object",
          "properties": {
            "sum": {
              "typeof": "function",
              "parameters": [
                {
                  "type": "number"
                },
                {
                  "type": "number"
                }
              ]
            },
            "prod": {
              "typeof": "function",
              "parameters": [
                {
                  "type": "number"
                },
                {
                  "type": "number"
                }
              ]
            }
          }
        },
        "objects": {
          "type": "object",
          "properties": {
            "length": {
              "typeof": "function",
              "parameters": [
                {
                  "type": "string"
                }
              ]
            },
            "extractAttribute": {
              "typeof": "function",
              "parameters": [
                {
                  "type": "object",
                  "properties": {},
                  "additionalProperties": true
                },
                {
                  "type": "string"
                }
              ]
            },
            "extractArrayEl": {
              "typeof": "function",
              "parameters": [
                {
                  "type": "array",
                  "items": {
                    "type": "number"
                  }
                },
                {
                  "type": "number"
                }
              ]
            }
          }
        }
      }
    }
  },
  "$schema": "http://json-schema.org/draft-07/schema#"
}
```

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -1006,11 +1006,10 @@ export class JsonSchemaGenerator {
     private getFunctionDefinition(funcType: ts.Type, definition: Definition): Definition {
         const node = funcType.getSymbol()!.getDeclarations()![0];
 
-        // tiberiu
         const parameters = (node as ts.FunctionLikeDeclaration).parameters;
         definition.typeof = "function";
         definition.parameters = parameters
-            .map((p) => this.tc.getTypeAtLocation(p.type))
+            .map((p) => this.tc.getTypeAtLocation(p.type!))
             .map((type) => this.getTypeDefinition(type));
 
         return definition;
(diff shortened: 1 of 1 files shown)


Tiberiu02:
@YousefED Do you think you could you take a look over this? It would be very useful for our project

domoritz:
How does this compare with what ts-json-schema-generator does?

Tiberiu02:
Didn't know about ts-json-schema-generator, let me take a look

PR #521, opened 2023-01-29
Add skiplibcheck as an option for TypeScript compiler

seren5240:
`skipLibCheck` is required for compatibility with our codebase to avoid TypeScript "Duplicate identifier" errors when running `typescript-json-schema`.


domoritz:
Thank you for the pull request.

PR #527, opened 2023-03-16
all unit test pass

netanel-mce:
The purpose of this PR is, to align all the unit tests that will pass successfully, so that later on we can write new features and make sure there is no regression


netanel-mce:
@YousefED @domoritz 
I updated the unit tests, so all unit tests pass now

domoritz:
I don't see a way to approve it. Also, please add a description.

netanel-mce:
> I don't see a way to approve it. Also, please add a description.

@domoritz Where do you want me to add a description?

domoritz:
There pull request should have a description

netanel-mce:
@domoritz done :)

test/schema.test.ts
```diff
@@ -45,7 +45,7 @@ export function assertSchema(
             settings.required = true;
         }
 
-        const files = [resolve(BASE + group + "/main.ts")];
```
domoritz, line 48: Huh, why is this correct?

netanel-mce: @domoritz 
the 'resolve' function return the full path, 
So we were not building the 'userSymbols' correctly, when the path is the full path
Because when we check whether a file is the user's file we check it against a relative path
see: https://github.com/YousefED/typescript-json-schema/blob/master/typescript-json-schema.ts#L1603 

Please run the "test" script locally before my branch and run after my branch for reproduce

domoritz:
Thanks. So weird, I still don't see the approve button.

netanel-mce:
@domoritz please try now, I marked your comment as resolved

domoritz:
Not sure what's up. Let's open a new pull request.

domoritz:
https://github.com/YousefED/typescript-json-schema/pull/530

Issue #516, opened 2022-11-29
Unsupported type `[keyof T]` 


Theartbug:
My team is trying to implement  the `AtLeastOne<T>` generic type created by the Microsoft Azure team: https://learn.microsoft.com/en-us/javascript/api/@azure/keyvault-certificates/requireatleastone?view=azure-node-latest

```
type RequireAtLeastOne<T> = { [K in keyof T]-?: Required<Pick<T, K>> & Partial<Pick<T, Exclude<keyof T, K>>>; }[keyof T]
```

We've poked around a bit and notice when we remove `[keyof T]` at the end, it can generate the `schemas.json`. We are wondering if it has something to do with recursion? 

The output error text is too long to retrieve full logs, but it looks like this:

```
....
aliasTypeArguments: [Array],
            constraintType: [TypeObject],
            resolvedApparentType: [Circular *24],
            modifiersType: [TypeObject]
          }
        },
        <ref *25> TypeObject {
          checker: {
            getNodeCount: [Function: getNodeCount],
            getIdentifierCount: [Function: getIdentifierCount],
            getSymbolCount: [Function: getSymbolCount],
            getTypeCount: [Function: getTypeCount],
            getInstantiationCount: [Function: getInstantiationCount],
            getRelationCacheSizes: [Function: getRelationCacheSizes],
            isUndefinedSymbol: [Function: isUndefinedSymbol],
            isArgumentsSymbol: [Function: isArgumentsSymbol],
            isUnknownSymbol: [Function: isUnknownSymbol],
            getMergedSymbol: [Function: getMergedSymbol],
            getDiagnostics: [Function: getDiagnostics],
            getGlobalDiagnostics: [Function: getGlobalDiagnostics],
            getRecursionIdentity: [Function: getRecursionIdentity],
            getUnmatchedProperties: [Function: getUnmatchedProperties],
            getTypeOfSymbolAtLocation: [Function: getTypeOfSymbolAtLocation],
            getTypeOfSymbol: [Function: getTypeOfSymbol],
            getSymbolsOfParameterPropertyDeclaration: [Function: getSymbolsOfParameterPropertyDeclaration],
            getDeclaredTypeOfSymbol: [Function: getDeclaredTypeOfSymbol],
            getPropertiesOfType: [Function: getPropertiesOfType],
            getPropertyOfType: [Function: getPropertyOfType],
            getPrivateIdentifierPropertyOfType: [Function: getPrivateIdentifierPropertyOfType],
[...]

PR #532, opened 2023-04-02
Support indexed access

netanel-mce:
**what is here**

- fix for the issue #516 
- unit test for single key and multi keys
- update api.md


domoritz (changes requested):
There are a lot of formatting changes. Can you reduce them so we can see the diff better?

netanel-mce:
> There are a lot of formatting changes. Can you reduce them so we can see the diff better?

@domoritz done

Issue #534, opened 2023-04-11
enums with single values should be "const"


lublak:
```ts
export interface Test {
    "testfield": "test"
}
```

i get a result like this:
```
"testfield": {
    "enum": [
        "test"
    ],
    "type": "string"
},
```

but it should look like this:

```
"testfield": {
    "const": "test",
    "type": "string"
},
```

PR #535, opened 2023-04-11
use const instead of enum

lublak:
Please:

fixes https://github.com/YousefED/typescript-json-schema/issues/534


domoritz:
Please remove the npm lock file

lublak:
@domoritz done

domoritz:
Tests are failing

lublak:
@domoritz can you try again

domoritz:
Why should we make this configurable? I feel it's a better default and would like it always enabled. What do you think?

lublak:
@domoritz then it is not compatible with your tests. (backwards compatible?)

domoritz:
Of we think this is better then we can release a new breaking version.

One question? How does this work if you have an enum with one value?

domoritz:
And speaking of tests, we need one.

lublak:
@domoritz the issue ist currenlty: The complete project doesn't work. If i just try npm test, the tests are currently failing (and no i have not really time to fix all issues).
Okay a breaking version can be fine for this.
If you have a enum with one value it works fine with most stuff.
But const is here a better solution. After all, that's what const is for.

domoritz:
You can just add one test and run that wither by marking the test as `.only` or commenting out the others.

lublak:
@domoritz i changed now the tests and its now a breaking change.

domoritz:
Please fix the tests so we can merge this.

lublak:
@domoritz yes!
The issue is, if i run my test local i can not see all issues directly. So i only can see if its fails here.
My local test looks like this:

grafik
grafik

but i hope i can fix it without testing it.

lublak:
@domoritz can you try again? :)

Issue #490, opened 2022-07-05
Invalid "undefined" type generated for `undefined` interface properties


rahil-p:
When an interface property is defined as `undefined`, the generated JSON schema produces `"type": "undefined"`, which is not a valid JSON schema type.

Source:

```typescript
interface Foo {
    bar: string;
}

interface FooWithoutId extends Foo {
    id?: undefined;
}
```

Expected:

```json
{
  "FooWithoutId": {
    "properties": {
      "bar": {
        "type": "string"
      }
    },
    "type": "object"
  }
}
```

Actual:

```json
{
  "FooWithoutId": {
    "properties": {
      "bar": {
        "type": "string"
      },
      "id": {
        "type": "undefined"
      }
    },
    "type": "object"
  }
}
```

<details>
  <summary>Full Example</summary>
  
  TypeScript:

```typescript
interface Foo {
    bar: string;
}

interface FooWithId extends Foo {
    id: string;
}

interface FooWithoutId extends Foo {
    id?: undefined;
}

export interface Schema {
    foo: FooWithId | FooWithoutId;
}
```

JSON Schema (generated):

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "definitions": {
    "FooWithId": {
      "properties": {
        "bar": {
          "type": "string"
        },
        "id": {
          "type": "string"
        }
      },
      "type": "object"
    },
    "FooWithoutId": {
      "properties": {
        "bar": {
          "type": "string"
        },
        "id": {
          "type": "undefined"
        }
      },
      "type": "object"
    }
  },
  "properties": {
    "foo": {
      "anyOf": [
        {
          "$ref": "#/definitions/FooWithId"
        },
        {
          "$ref": "#/definitions/FooWithoutId"
        }
      ]
    }
  },
  "type": "object"
}
```
  
</details>

PR #537, opened 2023-04-19
Compatibility to draft 07

netanel-mce:
**what is here**

- new flag `compatibility` to choose if output schema must to match to draft 07, if `draft-07` is chosen, so:
   - `undefined` is not a valid for `definition.type`, so: 
      - if main type is alias of undefined - we will throw error
      - if have a property undefined - we will ignore (like never)
   - `symbol` is not a valid for `definision.type`, so we write `object` instead
   - we not using in the operator `typeof` for `definition.type` only after check what returned from he
- new type `DefinitionTypeName` for save on `definition.type`
- unit test with compatibility `draft-07` and `none`
- updating README.md file
- updating api.md file
- fix issue: #490


domoritz:
I'm not following. Are you saying that we get errors if we generate non-standard schemas and that we currently do create those schemas? In that case I would strongly vote for making standard compliant output the default without a switch.

netanel-mce:
@domoritz Yes, that's the situation today, I'll remove the switch

netanel-mce:
> I'm not following. Are you saying that we get errors if we generate non-standard schemas and that we currently do create those schemas? In that case I would strongly vote for making standard compliant output the default without a switch.

@domoritz done, all schemas should compatibility to draft-07.

README.md
```diff
@@ -51,6 +51,7 @@ Options:
   --id                  Set schema id.                                               [string]  [default: ""]
   --defaultNumberType   Default number type.                                         [choices: "number", "integer"] [default: "number"]
   --tsNodeRegister      Use ts-node/register (needed for require typescript files).  [boolean] [default: false]
+  --compatibility       Output compatibility mode                                    [choices: "none", "draft-07"] [default: "none"]
```
domoritz, line 54: This is gone, right?

domoritz (approved):

README.md
netanel-mce: @domoritz right!, done.

domoritz:
Thank you

PR #552, opened 2023-06-18
Overcame typescript bug in property start with multi underscores

netanel-mce:
Please:
- [ V ] Make your pull request atomic, fixing one issue at a time unless there are many relevant issues that cannot be decoupled.
- [ V ] Provide a test case & update the documentation in the `Readme.md`

**what is here:**
as a continuation of #532 if one property start with two underscores or more, building schema are failed, 
it looks like a bug in typescript, have a 1 unnecessary underscore

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -1085,16 +1085,10 @@ export class JsonSchemaGenerator {
                     const typ = this.tc.getTypeAtLocation(indexSignature.type!);
                     let def: Definition | undefined;
                     if (typ.flags & ts.TypeFlags.IndexedAccess) {
-                        const targetName: string = (<any>clazzType).mapper?.target?.value;
+                        const targetName = ts.escapeLeadingUnderscores((<any>clazzType).mapper?.target?.value);
                         const indexedAccessType = <ts.IndexedAccessType>typ;
-                        const symbols: Map<string, ts.Symbol> = (<any>indexedAccessType.objectType).members;
-                        let targetSymbol = symbols?.get(targetName);
-
-                        // it looks like a bug in typescript,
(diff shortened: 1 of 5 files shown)


typescript-json-schema.ts
```diff
@@ -1088,7 +1088,13 @@ export class JsonSchemaGenerator {
                         const targetName: string = (<any>clazzType).mapper?.target?.value;
                         const indexedAccessType = <ts.IndexedAccessType>typ;
                         const symbols: Map<string, ts.Symbol> = (<any>indexedAccessType.objectType).members;
-                        const targetSymbol = symbols?.get(targetName);
+                        let targetSymbol = symbols?.get(targetName);
+
+                        // it looks like a bug in typescript,
+                        // if property start with 2 underscores and more, in objectType.members, have a 1 unnecessary underscore
```
domoritz, line 1094: Can you link to an issue in the ts repo so we know when we can remove this hack?

domoritz (approved):

typescript-json-schema.ts
netanel-mce: @domoritz https://github.com/microsoft/TypeScript/issues/54704

domoritz: Link from the comment I mean

netanel-mce: @domoritz so you want me to change the comment in the code, and you will merge this PR?

netanel-mce: @domoritz so, I switched to use in utility of typescript to convert to `escape` string

netanel-mce:
@domoritz Can you please release a new version for this?

domoritz:
DOne

Issue #559, opened 2023-07-23
How to generate title in human readable format?


FreePhoenix888:
# Code
```
export interface Abc {
  firstName: string
}
```
# Expected 
```
freephoenix888@freephoenix888:~/Programming/deep/firebase-push-notification$ typescript-json-schema ./src/push-notification.ts "Abc" --titles
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "properties": {
        "firstName": {
            "title": "First name",
            "type": "string"
        }
    },
    "type": "object"
}
```
# Actual
```
freephoenix888@freephoenix888:~/Programming/deep/firebase-push-notification$ typescript-json-schema ./src/push-notification.ts "Abc" --titles
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "properties": {
        "firstName": {
            "title": "firstName",
            "type": "string"
        }
    },
    "type": "object"
}
```

# Additional Info
Why do I need this: I want to use this generated JSON-Schema for form generation using [`react-jsonschema-form`](https://rjsf-team.github.io/react-jsonschema-form/)

FreePhoenix888:
@YousefED

FreePhoenix888:
@domoritz

FreePhoenix888:
We can use `capitalCase` from https://www.npmjs.com/package/case-anything

leqwasd:
Add JSDoc comment

```ts
export interface Abc {
    /**
     * First name
     */
    firstName: string;
}
```

acollazo25:
Remove `--titles` parameter from command and add `@title` JSDoc comment.

```ts
export interface Abc {
    /**
     * @title First name
     */
    firstName: string;
}
```

PR #561, opened 2023-07-24
Feature/capitalize title

FreePhoenix888:
Please:

Fixes https://github.com/YousefED/typescript-json-schema/issues/559

### DIFF
--- a/test/titles-option.test.ts
+++ b/test/titles-option.test.ts
@@ -1,0 +1,28 @@
+import { assert } from "chai";
+import { exec, getDefaultArgs,getProgramFromFiles, generateSchema, Definition } from "../typescript-json-schema";
+import {capitalCase} from 'case-anything'
+
+describe("out option", () => {
+    it("should have title in capital case", async () => {
+        try {
+            await exec("test/programs/interface-single/main.ts", "MyObject", {
+                ...getDefaultArgs(),
+                titles: true,
+            });
+            const program = getProgramFromFiles(["test/programs/interface-single/main.ts"]);
+            let schema = generateSchema(program, "MyObject", {titles: true}); 
+            assert.notEqual(schema, null)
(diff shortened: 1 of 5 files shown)


FreePhoenix888:
Fixes #559

FreePhoenix888:
@domoritz

test/titles-option.test.ts
```diff
@@ -0,0 +1,28 @@
...
+                ...getDefaultArgs(),
+                titles: true,
+            });
+            const program = getProgramFromFiles(["test/programs/interface-single/main.ts"]);
+            let schema = generateSchema(program, "MyObject", {titles: true}); 
+            assert.notEqual(schema, null)
+            let properties = schema!.properties;
+            assert.notEqual(schema!.properties, undefined)
+            properties = properties!;
+            for (const property of Object.entries(properties)) {
```
domoritz, line 18: Instead of testing with some logic, please just create a new test case or update an existing test case with a fully specified example.

domoritz (changes requested):

test/titles-option.test.ts
FreePhoenix888: Which?

domoritz: What’s your question?

FreePhoenix888: I do not understand what I should do

domoritz: Unit tests should have as little logic as possible to make them easier to debug and find issues. Your test uses the same library we use to generate the title so if something ever changes, we don’t notice it in the test. Ideally tests don’t have loops or call libraries.

domoritz:
Closing since there is some work to be done here to wrap this up. There is a workaround in https://github.com/YousefED/typescript-json-schema/issues/559#issuecomment-1962319791.

Issue #570, opened 2023-08-29
Add optional flag to convert const to single value enums


gassiss:
Hello! 👋 

We rely on this library to generate JSON schemas to use with OpenAPI. Unfortunately, [this PR](https://github.com/YousefED/typescript-json-schema/pull/535) introduced some breaking changes for us, as OpenAPI straight up doesn't allow constant values, and the workaround was to do exactly what the PR fixed.

Any chance we can get that behaviour back behind a feature flag? Seems like a pretty simple change, I have it working locally for our use case and would be happy to open a PR if that's OK.

Thanks for your work on this!

domoritz:
Happy to accept a pull request. Maybe we can add a flag for openapi. If I recall correctly, there were some other requests that you can search for.

gassiss:
nice! I'll open a PR for review soon. Thanks for the quick feedback!

Issue #287, opened 2019-05-29
Support openapi style JSON schema output


robaca:
In OpenAPI 3 documents, `"type": ["null", "string"]` is not supported. Instead you have to specify a property via `"type": "string", "nullable": true"`

Please provide a command line argument to optimize the output for OpenAPI 3 compatibility.

domoritz:
As far as I know, `nullable` is not a valid JSON schema property. Unless I am missing something, it is not possible to do this, unfortunately.

PR #571, opened 2023-08-30
add constAsEnum flag to convert constant values to single value enum

gassiss:
Closes https://github.com/YousefED/typescript-json-schema/issues/570

Please:


domoritz:
What do you think about https://github.com/YousefED/typescript-json-schema/issues/287? Should we have a flag for openapi compatibility instead?

gassiss:
> What do you think about #287? Should we have a flag for openapi compatibility instead?

In our use case, we never got any errors related to nullable strings, and we do have them. So looks like the ask is a no-op. Would you prefer if I changed the flag to `openApiCompatible` or something? (please suggest a better name, lol :sweat_smile: )

PR #590, opened 2024-02-13
fix: return overrides schema in getSchemaForSymbol

lundibundi:
Please:

Depends on https://github.com/YousefED/typescript-json-schema/pull/589.

The `getSchemaForSymbol` method doesn't return schemas overridden with `setSchemaOverride` and instead falls back to generating the type for that schema. This PR changes the behavior to always prioritize `setSchemaOverride`.

### DIFF
--- a/test/schema.test.ts
+++ b/test/schema.test.ts
@@ -168,6 +168,52 @@ describe("interfaces", () => {
             assert.deepEqual(schema.definitions!["MySubObject"], schemaOverride);
         }
     });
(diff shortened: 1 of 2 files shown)


domoritz:
Awesome. Thanks for the pull requests. Can you update this pull request so we can merge it?

lundibundi:
@domoritz done, thanks for quick merge.

When can I expect to have this released?

domoritz:
v0.63.0 is out with this fix included.

lundibundi:
@domoritz thank you 🎉

PR #593, opened 2024-03-14
Remove suppressImplicitAnyIndexErrors from tsconfig

raineorshine:
Please:

This PR is a continuation of the Typescript v5 upgrade (#564) that factors out the implicit use of index signatures, allowing `suppressImplicitAnyIndexErrors` and `ignoreDeprecations` to be removed from the tsconfig.

Due to the extensive use of indexing into interfaces (namely `Definition`), I have created a `DefinitionIndex` type that is a looser version of `Definition` allowing indexing. In other cases I needed to cast a type as a `Record` or cast a key `as keyof Definition` in order to allow the desired indexing.

Let me know if you know of a more elegant approach, or have any additional feedback.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,7 +9,6 @@
         "emitDecoratorMetadata": true,
         "declaration": true,
         "noImplicitAny": true,
-        "suppressImplicitAnyIndexErrors": true,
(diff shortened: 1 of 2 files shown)


domoritz:
Thank you

PR #595, opened 2024-04-03
fix: correctly identify imports with resolution-mode (#582)

### LINKED ISSUE #582: imports with resolution-mode not replaced
With my new TS config I have some imports in the output such as:

```
 "type": "object"
    },
    "import(\"apps/src/fileExportingTypes\",{assert:{\"resolution-mode\":\"import\"}}).MyType": {
      "anyOf": [
```

Fixing this regex make them disappear:

`var REGEX_FILE_NAME_OR_SPACE = /(\bimport\(".*?"(, \{ assert: \{ "resolution-mode": "(import|require)" \} \})?\)|".*?")\.| /g;`

FoxxMD:
Used fix provided by @jer-sen and closes #582 

 I _believe_ this only occurs for ESM projects with `"moduleResolution": "NodeNext"` in `tsconfig.json` and introduced in [TS 4.7](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#resolution-mode). This fix is backwards compatible.

### DIFF
--- a/typescript-json-schema.ts
+++ b/typescript-json-schema.ts
@@ -9,7 +9,7 @@ export { Program, CompilerOptions, Symbol } from "typescript";
 
 const vm = require("vm");
 
-const REGEX_FILE_NAME_OR_SPACE = /(\bimport\(".*?"\)|".*?")\.| /g;
+const REGEX_FILE_NAME_OR_SPACE = /(\bimport\(".*?"(, \{ assert: \{ "resolution-mode": "(import|require)" \} \})?\)|".*?")\.| /g;
 const REGEX_TSCONFIG_NAME = /^.*\.json$/;
 const REGEX_TJS_JSDOC = /^-([\w]+)\s+(\S|\S[\s\S]*\S)\s*$/g;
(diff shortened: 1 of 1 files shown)


typescript-json-schema.ts
```diff
@@ -9,7 +9,7 @@ export { Program, CompilerOptions, Symbol } from "typescript";
 
 const vm = require("vm");
 
-const REGEX_FILE_NAME_OR_SPACE = /(\bimport\(".*?"\)|".*?")\.| /g;
+const REGEX_FILE_NAME_OR_SPACE = /(\bimport\(".*?"(, \{ assert: \{ "resolution-mode": "(import|require)" \} \})?\)|".*?")\.| /g;
```
derdeka, line 12: Hi @FoxxMD,
thank you for pointing in the right direction. I had to use a different regex when using `typescript@22.14.0` and `typescript@5.8.3` with `esm` modules:

```suggestion
const REGEX_FILE_NAME_OR_SPACE = /(\bimport\(.*?\)|".*?")\.| /g;
```
I'm patching this line with the help of [patch-package](https://www.npmjs.com/package/patch-package) into `typescript-json-schema` to make it working again.

@YousefED Thank you for this great package! I would like to politely ask if this regular expression could be updated to work with ESM packages.

Issue #584, opened 2023-12-29
Support bigint


maartenbreddels:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

I used [patch-package](https://github.com/ds300/patch-package) to patch `typescript-json-schema@0.62.0` for the project I'm working on; it's an easier way to propose a fix than to open a PR.

I had your tool crash on a bigint, and this seems to solve my problem:

```diff
diff --git a/node_modules/typescript-json-schema/dist/typescript-json-schema.js b/node_modules/typescript-json-schema/dist/typescript-json-schema.js
index b230291..e6fb7af 100644
--- a/node_modules/typescript-json-schema/dist/typescript-json-schema.js
+++ b/node_modules/typescript-json-schema/dist/typescript-json-schema.js
@@ -482,6 +482,11 @@ var JsonSchemaGenerator = (function () {
                 definition.properties = {};
                 definition.additionalProperties = true;
             }
+            else if (propertyTypeString === "bigint") {
+                definition.type = "number";
+                definition.properties = {};
+                definition.additionalProperties = false;
+            }
             else {
                 var value = extractLiteralValue(propertyType);
                 if (value !== undefined) {
@@ -558,6 +563,7 @@ var JsonSchemaGenerator = (function () {
                         }
                     }
                 }
+
                 else {
                     var error = new TypeError("Unsupported type: " + propertyTypeString);
                     error.type = propertyType;
```

I hope this is helpful.

Regards,

Maarten

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

benmusson:
I had the same issue on 0.63.0, and this solution worked.

domoritz:
Please send a pull request if you'd like this feature supported.

domoritz:
Done in #596

PR #596, opened 2024-04-05
Support bigint

bradleypriest:
Pulls the change from #584 into a PR


typescript-json-schema.ts
```diff
@@ -705,6 +705,10 @@ export class JsonSchemaGenerator {
                 definition.type = "object";
                 definition.properties = {};
                 definition.additionalProperties = true;
+            } else if (propertyTypeString === "bigint") {
+                definition.type = "number";
+                definition.properties = {};
+                definition.additionalProperties = false;
```
domoritz, line 711:
```suggestion
                definition.additionalProperties = false;
```

Issue #336, opened 2019-12-11
Can not generate default value for using enum in class


Saul-Mirone:
```typescript
enum Fruit {
  Apple = 'apple',
  Orange = 'orange'
}
class Test {
  fruit: Fruit = Fruit.Apple;
}
```
Then I get:
```typescript
{
  fruit: {
    enum: ['apple', 'orange'],
    type: 'string',
  }
}
```
I expect the default can be set in here.

adampenn:
I am having this same issue. Any updated @Saul-Mirone

mohsin51:
having same issue default values are not being reflected in json-schema.

adiliqbl:
Facing the same issue. Any updates or workaround for this?

ahochsteger:
In addition to enums the handling of default values is broken for constants too.
While this produces the default value in the schema:
```typescript
class Test {
  num = 10;
}
```
This does not:
```typescript
const DEFAULT_NUM = 10;
class Test {
  num = DEFAULT_NUM;
}
```

ahochsteger:
The PR #600 implements support for referencing enums and constants as default values, would be great if I could get some feedback on it.
See `test/programs/default-properties-ref` for a test case.

PR #600, opened 2024-05-02
feat: support property defaults with constant and enum refs (#336)

ahochsteger:
* add helper functions getDeclarationValue() and getInitializerValue()
* make user symbols available to sandbox evaluation
* simplify initializer logic (eliminate ifs)

Please:

### DIFF
--- a/api.md
+++ b/api.md
@@ -667,6 +667,32 @@ class MyObject {
 ```
 
 
+## [default-properties-ref](./test/programs/default-properties-ref)
+
+```ts
+
+const defaultBooleanFalse = false;
+const defaultBooleanTrue = true;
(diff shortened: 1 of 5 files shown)


ahochsteger:
@YousefED it would be great if you could give me some feedback, if I took the right way to implement this feature. AFAICS it should be a bit more generic, but I'm not sure, if there may be some edge cases that are still not handled properly and should generate a warning.
The main change is, that I had to put the values of the user symbols into the sandbox to resolve constant/enum references.
In addition there are helper two functions to derive the value of the declarations and initializers. Additional logic can be easily added by switching the supported kinds.
All tests (including the new one at `test/programs/default-properties-ref`) are running successfully.

PR #603, opened 2024-05-11
fix: escape special characters in string template literals

### LINKED ISSUE #602: RegExp special characters are not escaped for string template literals
A string template literal like `${string}.${string}` currently produces a string pattern of `"^.*..*$"` in the resulting JSON schema while the correct pattern would be `"^.*\\..*$"`.

ahochsteger:
Please:

Fixes #602

### DIFF
--- a/api.md
+++ b/api.md
@@ -1445,6 +1445,24 @@ class MyObject {
 ```
 
 
+## [string-template-literal](./test/programs/string-template-literal)
+
+```ts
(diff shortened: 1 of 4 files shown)


domoritz:
The tests air failing.

ahochsteger:
@domoritz thanks for spotting, the quoting issue ist now fixed.

